PR #21, opened 2018-10-08
Enable real TPM usage in this sample code.

yphuangms:
Add compiler symbol "_USE_TPMSIMULATOR" at build time to switch between TPM simulator and real TPM access.
Use device hostname as registration Id.

## Purpose

* Create a compiler symbol to switch between "TPM Simulator" code and "real TPM" code makes the code more available for people who would like to use "real TPM". 

* Replace constant registration Id with hostname of device.
Use hostname of device as registration Id. This illustrate a scenario that registration Id could be something uniquely represent this device and stored in device, this could also be serial number.  

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/provisioning/Samples/device/TpmSample/Program.cs
+++ b/provisioning/Samples/device/TpmSample/Program.cs
@@ -1,11 +1,21 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+// Comment this define to switch access to real TPM
+#define _USE_TPMSIMULATOR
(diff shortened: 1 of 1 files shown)


CIPop (approved):
Thanks @yphuangms !

PR #42, opened 2019-02-25
Removing timeouts and looping forever in device streams proxy samples.

rezasherafat:
## Purpose

* Remove timeouts in the device and service local proxy
* Make device local proxy loop forever (still some error conditions are not handled)

```
[ ] Yes
[ X ] No
```

## Pull Request Type

```
[ X ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code
Follow instructions [here](https://docs.microsoft.com/en-us/azure/iot-hub/quickstart-device-streams-proxy-csharp#ssh-to-a-device-via-device-streams)

* Test the code

```
```

## What to Check
* Timeouts removed
* Device local proxy loops forever

### DIFF
--- a/iot-hub/Quickstarts/device-streams-proxy/device/DeviceStreamSample.cs
+++ b/iot-hub/Quickstarts/device-streams-proxy/device/DeviceStreamSample.cs
@@ -51,12 +51,23 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
         public async Task RunSampleAsync()
         {
-            await RunSampleAsync(true).ConfigureAwait(false);
+            while (true)
+            {
+                try
+                {
+                    await RunSampleAsync(true).ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    Console.WriteLine("Got an exception: {0}", ex);
+                }
+                Console.WriteLine("Waiting again...");
+            }
         }
 
         public async Task RunSampleAsync(bool acceptDeviceStreamingRequest)
         {
-            using (var cancellationTokenSource = new CancellationTokenSource(new TimeSpan(0, 5, 0)))
+            using (var cancellationTokenSource = new CancellationTokenSource())
             {
                 DeviceStreamRequest streamRequest = await _deviceClient.WaitForDeviceStreamRequestAsync(cancellationTokenSource.Token).ConfigureAwait(false);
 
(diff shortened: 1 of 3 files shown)


rezasherafat:
@ewertons, @CIPop to review. note that this does not handle concurrent connections but enable RDP scenario which had regressed due to lack of support for back-to-back streaming connections on the proxy.

rezasherafat:
@vishal-gandhi as FYI

iot-hub/Quickstarts/device-streams-proxy/device/Program.cs
```diff
@@ -31,8 +31,8 @@ public static class Program
         private static string s_port = Environment.GetEnvironmentVariable("REMOTE_PORT");
 
         // Select one of the following transports used by DeviceClient to connect to IoT Hub.
-        private static TransportType s_transportType = TransportType.Amqp;
-        //private static TransportType s_transportType = TransportType.Mqtt;
+        //private static TransportType s_transportType = TransportType.Amqp;
```
ewertons: Need to confirm if C# SDK team has a standard of having AMQP as the default option.
Are you changing this on purpose?

iot-hub/Quickstarts/device-streams-proxy/device/DeviceStreamSample.cs
```diff
@@ -51,12 +51,23 @@ private static async Task HandleOutgoingDataAsync(NetworkStream localStream, Cli
 
         public async Task RunSampleAsync()
         {
-            await RunSampleAsync(true).ConfigureAwait(false);
+            while (true)
```
ewertons: I think the best way is to:
- expose this cancellation token as an argument of RunSampleAsync, and
- have the infinite loop but check if the cancellation token has been triggered;
- on Main invoke RunSampleAsync passing a cancellation token, but without a timer.

that way the sample becomes safer in case a customer gets it as is to use in production, having a way to escape the infinite loop.

But it would be good to get some input from @CIPop

iot-hub/Quickstarts/device-streams-proxy/device/Program.cs
rezasherafat: If anything, I prefer using MQTT as default. MQTT is the protocol we have repeatedly heard from customers to have more support for.

iot-hub/Quickstarts/device-streams-proxy/device/DeviceStreamSample.cs
rezasherafat: @ewertons, done. Have a look in the new committed version

iot-hub/Quickstarts/device-streams-proxy/device/Program.cs
rezasherafat: Done

PR #59, opened 2019-06-01
Need to close  the stream  and improved buffer code.

djaus2:
Inserted stream socket closure code in service as per the device as the device reports incorrect closure at this end.

Improved buffer coding in service in line with that in device. Required for .Net Standard compilation.

## Purpose
First change removes an error/warning that can occur. The second tightens up the code.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[X] Refactoring (no functional changes, no api changes)
[X] Other... Please describe:
Enables code to be built as a .Net Standard library, when refactored


## How to Test
*  Get the code

```
git clone https://github.com/djaus2/AziothubDeviceStreaming.git
cd AziothubDeviceStreaming
git checkout Master
npm install
```

* Test the code
Set target to Any CPU and configure to run multiple projects 
- DNCore_OrigEchoSample_DeviceClientStreaming  and
- DNCore_OrigEchoSample_ServiceClientStreaming
Need to create an IoT Hub and add settings to AzureConnections.cs in Solution common folder (or in DNCore_AzureConnections project)

## What to Check
Consoles should display that the Service passes a message, the device reads it and returns it.. And the service gets it back. In my project the device uppercases the message.

## Other Information
You might like to try the refactored projects:
DNCore_DeviceApp and  DNCore_ServiceApp


CIPop:
@ewertons PTAL

PR #72, opened 2019-08-15
Update 'rules' in scripts to 'queries'.

robinsh:
Replaces "rules" in the scripts with the word "queries" to be accurate and match the articles that use these scripts.


robinsh:
#sign-off

PR #78, opened 2019-09-17
DevOps: YAML changes

CIPop:
- Adding device connection string variable
- Adding cred scan overrides.


CIPop:
Samples build failed as our hub has too many devices/registrations.

PR #82, opened 2019-09-30
Fix storage conn string var.

robinsh:
Change the variable for the storage connection string (remove csharp). 
Comment out code they shouldn't accidentally run in the importexportdevicessample.


CIPop:
@abhipsaMisra PTAL in case this fails within the cleanup process. We can now use @robinsh 's sample to clean the entire hub (the only code change may be to ignore the "DoNotDelete" devices).

PR #53, opened 2019-03-27
Allow for parallel proxied connections.

rezasherafat:
## Purpose

* Allow for concurrent streaming connections in device proxy sample

## Does this introduce a breaking change?
```
[ ] Yes
[X] No
```

## Pull Request Type
What kind of change does this Pull Request introduce?
Uses un-awaited tasks to process parallel streaming connections.

```
[ ] Bugfix
[X] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
*  Run [device and service proxies as before](https://docs.microsoft.com/en-us/azure/iot-hub/quickstart-device-streams-proxy-csharp#ssh-to-a-device-via-device-streams).

## What to Check
Verify that the following are valid
* Use of device streams in independent async tasks.

## Other Information
This change is per offline discussions with @ewertons.

Issue #38, opened 2019-02-06
Exception after first login with device streams proxy sample


dsajanice:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
>
1. Use device streams proxy samples to set up device and service side SSH proxies on the SAME machine (experiencing different issues cross machine).

2. Log in using SSH via service side proxy. See it work.

3. Exit SSH session.

4. Log in again using SSH via service side proxy. This second login throws an exception on the device proxy:

Microsoft.Azure.Devices.Common.Exceptions.DeviceNotFoundException: Device {"Message":"{\"errorCode\":404001,\"trackingId\":\"c45f6aee9d2b4cc78402ad05811ca501-G:11-TimeStamp:02/06/2019 22:41:10\",\"message\":\"Device unavailable\",\"timestampUtc\":\"2019-02-06T22:41:10.9080359Z\"}","ExceptionMessage":""} not registered
   at Microsoft.Azure.Devices.HttpClientHelper.ExecuteAsync(HttpClient httpClient, HttpMethod httpMethod, Uri requestUri, Func`3 modifyRequestMessageAsync, Func`2 isMappedToException, Func`3 processResponseMessageAsync, IDictionary`2 errorMappingOverrides, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.HttpClientHelper.PostAsync[T,T2](Uri requestUri, T entity, TimeSpan operationTimeout, IDictionary`2 errorMappingOverrides, IDictionary`2 customHeaders, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.AmqpServiceClient.CreateStreamAsync(Uri uri, DeviceStreamRequest deviceStreamRequest, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Samples.DeviceStreamSample.HandleIncomingConnectionsAndCreateStreams(String deviceId, ServiceClient serviceClient, TcpClient tcpClient) in /home/jadsa-auto/Repos/azure-iot-samples-csharp/iot-hub/Quickstarts/device-streams-proxy/service/DeviceStreamSample.cs:line 60
[...]

PR #93, opened 2019-12-17
Cancel the remote stream if one is already open.

aguywithcode:
## Purpose
* For the device streams local proxy sample. If you try to open a stream before the existing steam is closed, it errors out. This fix sends a cancel request to the existing stream before trying to open a new one.

[ ] Yes
[x] No
```

## Pull Request Type

<!-- Please check the one that applies to this PR using "x". -->
```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
```
git clone git@github.com:aguywithcode/azure-iot-samples-csharp.git
cd azure-iot-samples-csharp
git checkout devicestreams-local-proxy-fix
cd iot-hub/Quickstarts/device-streams-proxy/
```

* Test the code
*  Run the local and device proxies
*  SSH to the local proxy
*  Exit the SSH
*  SSH again *before* the stream has closed

## What to Check
*  Stream should be properly closed and new stream created for the new client

## Other Information
<!-- Add any other helpful information that may be needed here. -->

### DIFF
--- a/iot-hub/Quickstarts/device-streams-proxy/service/DeviceStreamSample.cs
+++ b/iot-hub/Quickstarts/device-streams-proxy/service/DeviceStreamSample.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Azure.Devices.Samples
 {
     public class DeviceStreamSample
     {
+        private CancellationTokenSource _cancellationTokenSource;
         private ServiceClient _serviceClient;
         private String _deviceId;
         private int _localPort;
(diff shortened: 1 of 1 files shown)


aguywithcode:
I *believe* this addresses #38 

Also this cancels the first stream it doesn't allow for multiple clients to connect simultaneously. There is a PR (#53) that handles multiple connections on the device proxy. The local proxy could be modified to handle multiple client streams. But for now I just want it not to break when trying to re-establish an ssh connection.

PR #99, opened 2020-01-06
Update digital twin samples to support net core 3.0

timtay-microsoft:
Previously only supported net core 2.2, but now supports both 2.2 and 3.0

### DIFF
--- a/digitaltwin/Samples/device/EnvironmentalSensorSample/ModelDefinitionInterface.cs
+++ b/digitaltwin/Samples/device/EnvironmentalSensorSample/ModelDefinitionInterface.cs
@@ -35,11 +35,11 @@ namespace EnvironmentalSensorSample
             }
         }
 
-                              /// <summary>
-                              /// Callback on command received.
-                              /// </summary>
-                              /// <param name="commandRequest">information regarding the command received.</param>
-                              /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        /// <summary>
+        /// Callback on command received.
+        /// </summary>
+        /// <param name="commandRequest">information regarding the command received.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
         protected override Task<DigitalTwinCommandResponse> OnCommandRequest(DigitalTwinCommandRequest commandRequest)
         {
             // There is only one command that ModelDefinition defines, and it is getModelDefinition. That command must specify the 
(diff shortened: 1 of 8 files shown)


digitaltwin/Samples/device/EnvironmentalSensorSample/ModelDefinitionInterface.cs
```diff
@@ -38,11 +49,11 @@ protected override async Task<DigitalTwinCommandResponse> OnCommandRequest(Digit
                 string commandPayload = JsonConvert.DeserializeObject<string>(commandRequest.Payload);
                 if (commandPayload.Equals(EnvironmentalSensorInterface.EnvironmentalSensorInterfaceId))
                 {
-                    return new DigitalTwinCommandResponse(StatusCodeCompleted, this.environmentalSensorModelDefinition);
+                    return Task<DigitalTwinCommandResponse>.Factory.StartNew(() => new DigitalTwinCommandResponse(StatusCodeCompleted, this.environmentalSensorModelDefinition));
```
drwill-ms: This should just be:

return Task.FromResult(new DigitalTwinCommandResponse(StatusCodeCompleted, this.environmentalSensorModelDefinition));

digitaltwin/Samples/device/EnvironmentalSensorSample/ModelDefinitionInterface.cs
```diff
@@ -18,18 +19,28 @@ public class ModelDefinitionInterface : DigitalTwinInterfaceClient
...
+        private const string EnvironmentalSensorInterfaceFileAssemblyName = "EnvironmentalSensorSample";
+
         public ModelDefinitionInterface(string interfaceInstanceName)
             : base(ModelDefinitionInterfaceId, interfaceInstanceName)
         {
-            environmentalSensorModelDefinition = File.ReadAllText("../../../EnvironmentalSensor.interface.json");
+            var assembly = Assembly.GetExecutingAssembly();
+            var assemblyResourceName = EnvironmentalSensorInterfaceFileAssemblyName + "." + EnvironmentalSensorInterfaceFileName;
+
+            using (Stream stream = assembly.GetManifestResourceStream(assemblyResourceName))
```
drwill-ms: Consider using the new syntax:

using Stream stream = assembly.GetManifestResourceStream(assemblyResourceName);
using StreamReader reader = new StreamReader(stream);

No need for code blocks anymore. When the variables go out of scope, Dispose will be called.

timtay-microsoft: Cool, I didn't know that was an option!

digitaltwin/Samples/device/EnvironmentalSensorSample/ModelDefinitionInterface.cs
timtay-microsoft: Sure thing

PR #107, opened 2020-02-21
Add UTF-32 encoding.

robinsh:
Change the encoding to use UTF-32 encoding, and add a method to decode the messages for testing.


robinsh:
Added UTF-encoding for the body of the message. Added method the customer can use to decode the body so they can view it. 
#sign-off

PR #110, opened 2020-03-25
Service samples use best practices and dispose

drwill-ms:


iot-hub/Samples/service/DeviceStreamingSample/ServiceClientStreamingSample.csproj
```diff
@@ -1,8 +1,9 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
bikamani: why 2.1 specifically?

iot-hub/Samples/service/EdgeDeploymentSample/EdgeDeploymentSample.cs
```diff
@@ -80,29 +80,29 @@ public async Task RunSampleAsync()
...
             for (var i = 0; i < numToAdd; i++)
             {
-                var deviceName = $"{deviceIdPrefix}_{i:D8}-{Guid.NewGuid().ToString()}";
+                string deviceName = $"{deviceIdPrefix}_{i:D8}-{Guid.NewGuid().ToString()}";
                 var device = new Device(deviceName)
                 {
                     Capabilities = new DeviceCapabilities
                     {
-                        IotEdge = true
+                        IotEdge = true,
```
bikamani: I have seen you usually adding "," for initializers like this one. Just wondering why?

iot-hub/Samples/service/DeviceStreamingSample/ServiceClientStreamingSample.csproj
drwill-ms: Just making them all the same, and 2.1 is an LTS.

drwill-ms: https://dotnet.microsoft.com/download/dotnet-core

iot-hub/Samples/service/EdgeDeploymentSample/EdgeDeploymentSample.cs
drwill-ms: It is a good practice to follow, and C# allows trailing commas.

1. When you need to add another property, the last property listed doesn't need to change, because it already has a comma.
1. If you want to remark out or reorder any of the properties, none of the other lines need to change.

iot-hub/Samples/service/CleanUpDevicesSample/CleanUpDevicesSample.csproj
```diff
@@ -1,12 +1,13 @@
...
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.1</TargetFramework>
+    <LangVersion>8</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Devices" Version="1.*" />
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.20.0" />
```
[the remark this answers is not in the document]
drwill-ms: How could they possibly be better? They are indeterministic. This is not a pattern for success. No one should do this in their prod app, right?

iot-hub/Samples/service/ServiceClientSample/ServiceClientSample.csproj
```diff
@@ -1,12 +1,13 @@
...
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.1</TargetFramework>
+    <LangVersion>8</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
-  	<PackageReference Include="Microsoft.Azure.Devices" Version="1.*" />
```
bikamani: what version was it picking up if it was "*"? 

& so does this mean we change references in sample project with each release?

drwill-ms: Whatever the latest was for that person when they built the project.

We don't have to rev our samples' references on each release. Each release we have stands the test of time for a while. We can choose to update it on occasion if we have new API surface that we want to illustrate, though.

iot-hub/Samples/service/CleanUpDevicesSample/CleanUpDevicesSample.csproj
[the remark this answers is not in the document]
abhipsaMisra: For a production app, sure; you would want to take a dependency on a particular package. 
For a sample demonstrating code usage, if we take a dependency on a particular version, we would have to keep updating the sample any time that we release a new nuget package. Floating version there would ensure that the latest package gets picked up from the feed.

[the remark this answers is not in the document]
abhipsaMisra: What I am focusing here is that sample is demonstrating code usage, and not nuget package usage/ project configuration. I don't want to put the wrong sample out there just because it makes our work easier; but what I am saying is that the customer is looking at our code and not csproj configuration.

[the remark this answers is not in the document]
drwill-ms: Right, and I want our samples to follow these production practices.

We don't have to update to the latest package version every time we ship. I suggest we update to a latest version any time we make use of any new API surface. Otherwise the referenced release is still good and users can use it.

I think our release version advances should be explicit and chosen, not automatically selected without testing validation.

iot-hub/Samples/service/ServiceClientSample/ServiceClientSample.csproj
bikamani: yes but then we need to remember to do it. One extra step to do for releases. We should at least add it in our release notes.

drwill-ms: We don't need to update it for each release. There is no such requirement.

PR #115, opened 2020-04-11
Update pipeline config, build script to run cleanup first

abhipsaMisra:


iot-hub/Samples/service/CleanUpDevicesSample/CleanUpDevicesSample.cs
```diff
@@ -83,50 +50,52 @@ public async Task RunCleanUpAsync()
                     _devicesToDelete.RemoveRange(0, i);
```
drwill-ms, line 50: I question how well this will work. We're incrementing i, but also removing those items from the collection.

If I were to write this, I'd go from the end to the beginning, and remove as I go. That way the index of the next batch of items don't get impacted by removals.

iot-hub/Samples/service/CleanUpDevicesSample/CleanUpDevicesSample.cs
```diff
@@ -83,50 +50,52 @@ public async Task RunCleanUpAsync()
...
+            Console.WriteLine("Query devices for cleanup:");
+            string sqlQueryString = "select * from devices";
+            IQuery query = _rm.CreateQuery(sqlQueryString, QueryBatchSize);
 
-                Console.WriteLine($"-- Total # of devices deleted: {devicesDeleted}");
-            
-                var configurations = await _rm.GetConfigurationsAsync(100).ConfigureAwait(false);
+            while (query.HasMoreResults)
+            {
+                IEnumerable<Shared.Twin> result = await query.GetNextAsTwinAsync();
```
drwill-ms, line 86: configure await

iot-hub/Samples/service/CleanUpDevicesSample/CleanUpDevicesSample.cs
abhipsaMisra: We are removing these items from the collection once we're done incrementing i and adding those items to another collection. One problem I see here is if the registry manage operation fails, we've removed the items from our primary list, but they haven't been deleted. So I am moving the remove operation after registry manager operation.

PR #119, opened 2020-05-15
[Samples] Migrate "Read device-to-cloud" to the new Event Hubs client

jsquire:
## Purpose

The focus of these changes is to update the version of the Event Hubs client library in use for the sample, moving to the new `Azure.Messaging.EventHubs` package and enhancing the documentation which accompanies it.

- [x] Yes
- [ ] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Documentation content changes
- [x] Other... Please describe: `Dependency update`

### Obtain the code
```
git clone git@github.com:jsquire/azure-iot-samples-csharp.git ./squire-samples
cd squire-samples
git checkout sample/d2c-update
dotnet restore
dotnet build
```

### Run the sample
```
- Add your event hub compatible endpoint information by running Azure CLI
- Update the code in `<root>`\iot-hub\Quickstarts\read-d2c-messages\ReadDeviceToCloudMessages.cs`. 
- Send data to your IotHub via any means you are familiar with
- Run the `read-d2c-messages` project and observe the console window
```

## What to Check
```
- Verify that the sample prints out messages that you have sent
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,9 @@ azure-iot-samples-csharp provides a set of easy-to-understand, continuously-test
 
 ## Prerequisites
 
-- .NET Core SDK 2.1.0 or greater on your development machine.  You can download the .NET Core SDK for multiple platforms from [.NET](https://www.microsoft.com/net/download/all).  You can verify the current version of C# on your development machine using 'dotnet --version'.
+- .NET Core SDK 3.0.0 or greater on your development machine.  You can download the .NET Core SDK for multiple platforms from [.NET](https://www.microsoft.com/net/download/all).  You can verify the current version of C# on your development machine using 'dotnet --version'.
+
+  **Note:** The samples can be compiled using the NET Core SDK 2.1 SDK if the language version of projects using C# 8 features are changed to `preview`.
 
 
 ## Resources
--- a/iot-hub/Quickstarts/read-d2c-messages/README.md
+++ b/iot-hub/Quickstarts/read-d2c-messages/README.md
@@ -1,0 +1,58 @@
+# Read device-to-cloud messages
+
+This sample demonstrates how to use the Azure Event Hubs client library for .NET to read messages sent from a device by using the built-in Event Hub that exists by default for every IoT Hub instance. 
+
+## Prerequisites
+
+The .NET Core SDK 3.0.0 or greater is recommended.  You can download the .NET Core SDK for multiple platforms from [.NET](https://www.microsoft.com/net/download/all).  You can verify the current version of C# on your development machine using 'dotnet --version'.
+
+This sample can also be compiled using the NET Core SDK 2.1 SDK if the language version of project is changed to `preview`.
+
+## Obtain the Event Hubs-compatible connection string
+
+You can get the Event Hubs-compatible connection string to your IotHub instance via the Azure portal or
+by using the Azure CLI.
+
+If using the Azure portal, see [Built in endpoints for IotHub](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-read-builtin#read-from-the-built-in-endpoint) to get the Event Hubs-compatible connection string and assign it to the variable `connectionString` in the sample. You can skip the Azure CLI instructions in the sample after this.
+
+If using the Azure CLI, you will need to run the below before running this sample to get the details required to form the Event Hubs compatible connection string:
+
+```bash
+az iot hub show --query properties.eventHubEndpoints.events.endpoint --name {your IoT Hub name}
+az iot hub show --query properties.eventHubEndpoints.events.path --name {your IoT Hub name}
+az iot hub policy show --name service --query primaryKey --hub-name {your IoT Hub name}
+```
+
+If you can do neither of the above and need to programmatically get this information, the sample [How to request the IoT Hub built-in Event Hubs-compatible endpoint connection string](https://github.com/Azure/azure-sdk-for-net/blob/master/samples/iothub-connect-to-eventhubs/README.md) demonstrates how to do so.
+
+## WebSocket and proxy support
+
+If you would like to use WebSockts, with our without a proxy, you will need to create a set of options for the [`EventHubConsumerClient`](https://docs.microsoft.com/en-us/dotnet/api/azure.messaging.eventhubs.consumer.eventhubconsumerclient?view=azure-dotnet) to configure its behavior.  Proxy support is offered via the [`IWebProxy`](https://docs.microsoft.com/dotnet/api/system.net.iwebproxy?view=netcore-3.1) interface, which includes the built-in [`WebProxy`](https://docs.microsoft.com/dotnet/api/system.net.webproxy?view=netcore-3.1) class.  Any proxy must be explicitly passed; the client does not assume that any proxy set via the ambient environment or system-wide is desired.
+
+The options may be created as follows:
+
+```csharp
+var options = new EventHubConsumerClientOptions();
+
+// This line sets the transport to use WebSockets.
+options.ConnectionOptions.TransportType = EventHubsTransportType.AmqpWebSockets;
+
+// The following lines configure the options for proxy use.
+IWebProxy proxy = new WebProxy("<< URI TO PROXY >>", true);
+options.ConnectionOptions.Proxy = proxy;
+```
+
+Once you have your options, you'll need to pass them to the client constructor.  Each constructor accepts a set of options as the last parameter, such as:
+
+```csharp
+string consumerGroup = EventHubConsumerClient.DefaultConsumerGroupName;
+await using var consumer = new EventHubConsumerClient(consumerGroup, "<< CONNECTION STRING >>", "<< EVENT HUB >>", options);
+```
+
+## Additional Resources
+
+- [Event Hubs Product Documentation](https://docs.microsoft.com/azure/event-hubs/)
+- [Event Hubs Client Library Documentation](https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/eventhub/Azure.Messaging.EventHubs/README.md)
+- [Event Hubs Samples](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/eventhub/Azure.Messaging.EventHubs/samples/README.md)
+- [Event Processor Client Library Documentation](https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/eventhub/Azure.Messaging.EventHubs.Processor/README.md)
+- [Event Processor Samples](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/eventhub/Azure.Messaging.EventHubs.Processor/samples/README.md)
--- a/iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
+++ b/iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
@@ -1,101 +1,111 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-// This application uses the Microsoft Azure Event Hubs Client for .NET
-// For samples see: https://github.com/Azure/azure-event-hubs/tree/master/samples/DotNet
-// For documenation see: https://docs.microsoft.com/azure/event-hubs/
+// This application uses the Azure Event Hubs Client for .NET
+// For samples see: https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/eventhub/Azure.Messaging.EventHubs/samples/README.md
+// For documentation see: https://docs.microsoft.com/azure/event-hubs/
+
 using System;
-using Microsoft.Azure.EventHubs;
-using System.Threading.Tasks;
-using System.Threading;
 using System.Text;
-using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Messaging.EventHubs.Consumer;
 
 namespace read_d2c_messages
 {
-    class ReadDeviceToCloudMessages
+    public class ReadDeviceToCloudMessages
     {
         // Event Hub-compatible endpoint
         // az iot hub show --query properties.eventHubEndpoints.events.endpoint --name {your IoT Hub name}
-        private readonly static string s_eventHubsCompatibleEndpoint = "{your Event Hubs compatible endpoint}";
+        private const string EventHubsCompatibleEndpoint = "{your Event Hubs compatible endpoint}";
 
         // Event Hub-compatible name
         // az iot hub show --query properties.eventHubEndpoints.events.path --name {your IoT Hub name}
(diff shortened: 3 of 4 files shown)


jsquire:
This is similar to what was done to the Node sample in Azure-Samples/azure-iot-samples-node#39

//cc @elhorton, @jebrando, @JimacoMS3, @robinsh, @philmea, @ramya-rao-a

jsquire:
@JimacoMS3:  

I updated the members referenced in the [Quickstart documentation](https://docs.microsoft.com/en-us/azure/iot-hub/quickstart-send-telemetry-dotnet#read-the-telemetry-from-your-hub).  I noticed an omission for specifying the IoT SAS key name in the current write-up, which would require editing the docs to fix. 
 Since an edit was needed, I adjusted the existing members that were not using idiomatic C# to conform to expected norms.   

The set now looks like:
- `EventHubsCompatibleEndpoint` : documented as `s_eventHubsCompatibleEndpoint`
- `EventHubName` : documented as `s_eventHubsCompatiblePath`
- `IotHubSasKey`: documented as `s_iotHubSasKey`
- `IotHubSasKeyName`: previously undocumented.   Appeared as `s_iotHubSasKeyName` in the code.

README.md
```diff
@@ -16,7 +16,9 @@ azure-iot-samples-csharp provides a set of easy-to-understand, continuously-test
 
 ## Prerequisites
 
-- .NET Core SDK 2.1.0 or greater on your development machine.  You can download the .NET Core SDK for multiple platforms from [.NET](https://www.microsoft.com/net/download/all).  You can verify the current version of C# on your development machine using 'dotnet --version'.
+- .NET Core SDK 3.0.0 or greater on your development machine.  You can download the .NET Core SDK for multiple platforms from [.NET](https://www.microsoft.com/net/download/all).  You can verify the current version of C# on your development machine using 'dotnet --version'.
```
jsquire, line 19: Please let me know if this causes an issue;  I'd prefer not to set the language version to `preview` if possible, but its an option if needed.

JimacoMS3: @jsquire Do you mean going from .NET Core SDK 2.1 -> 3.0? I'm not sure @jebrando ?

iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
```diff
@@ -1,101 +1,108 @@
...
+        // az iot hub show --query properties.eventHubEndpoints.events.endpoint --name {your IoT Hub name}
+        private const string EventHubsCompatibleEndpoint = "{your Event Hubs compatible endpoint}";
+
+        // Event Hub-compatible name
+        // az iot hub show --query properties.eventHubEndpoints.events.path --name {your IoT Hub name}
+        private const string EventHubName = "{your Event Hubs compatible name}";
+
+        // az iot hub policy show --name service --query primaryKey --hub-name {your IoT Hub name}
+        private const string IotHubSasKey = "{your service primary key}";
+        private const string IotHubSasKeyName = "{your service primary key's name}";
```
JimacoMS3, line 28: @jebrando Can you please just set this constant to "service". That's the policy that we're getting in the CLI command.

README.md
jsquire: Yes;  with Core 3.1 being LTS and the recommended version, I believe it makes sense to bump.  With the track two libraries built on C# 8 concepts, it also helps to avoid using `preview` as the language version or making samples more complicated by manually unrolling asynchronous enumerators.

iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
```diff
@@ -1,101 +1,108 @@
...
+{
+    class ReadDeviceToCloudMessages
+    {
+        // Event Hub-compatible endpoint
+        // az iot hub show --query properties.eventHubEndpoints.events.endpoint --name {your IoT Hub name}
+        private const string EventHubsCompatibleEndpoint = "{your Event Hubs compatible endpoint}";
+
+        // Event Hub-compatible name
+        // az iot hub show --query properties.eventHubEndpoints.events.path --name {your IoT Hub name}
+        private const string EventHubName = "{your Event Hubs compatible name}";
```
JimacoMS3, line 24: Not sure why you made this constant **EventHubName** rather than **EventHubsCompatiblePath**. Is there a reason? (I'm not arguing -- just asking.)

iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
jsquire: Not sure if you intended to reference me... but, if so, I'm not sure that I follow the ask.  Would you be so kind as to clarify?

iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
jsquire: Yes; the concept in the Event Hubs client across each of the languages in track two  is called `EventHubName`.   The `EntityPath` semantics were largely a holdover from the days of Event Hubs and Service Bus sharing a neutral interface where a queue/topic/subscription/event hub could be specified.

iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
JimacoMS3: Yup please set the value to "service". 

IoT Hub exposes a bunch of different shared access policies that grant different privileges, i.e. iothubowner, service, device, registryRead, registryReadWrite. We want people to use the least privileged policy for whatever ops they are wanting to perform. In this case, that would be the "service" policy. I'm honestly not really sure how/whether that has any effect on the privileges granted by EH. I think I started a thread at one point, but dropped it.

README.md
JimacoMS3: I agree with what you're saying, but don't really have enough personal knowledge to say what we should do here. 

@jebrando @elhorton I think the device SDK still specifies ,NET Core 2.1 [Azure IoT Device SDKs Platform Support](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-device-sdk-platform-support#microsoft-sdks-and-device-platform-support) (Scroll down to **.NET SDK**). I don't know whether the .NET Service SDK similarly specifies .NET Core 2.1 and whether introducing .NET Core 3.1 here is going to be confusing for customers.  Can one of you please help out here?

iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
JimacoMS3: Thanks for the context! That doesn't seem worth arguing about -- **EventHubName** it is. :)

README.md
elhorton: I can't speak to the C# SDK but the Platform Support doc is up to date. Adding @sharmasejal and @prmathur-microsoft to help from the C# side

JimacoMS3:
Added this docs PR for required changes to quickstart: https://github.com/MicrosoftDocs/azure-docs-pr/pull/115441#issuecomment-629529551. 

So far, I've changed the names of the constants as you suggested. (Note: I'm not adding the service name one to the table as it should just be set to "service".) Ill wait to see where we land on .NET Core 2.1 v 3.1 before changing the Prerequisites.

README.md
jsquire: One additional bit of context: this impacts only the recommended SDK for compiling; it does not imply that a new runtime would be needed for deploying.  I've left the target framework for the updated sample as `netcoreapp2.1`.

Apologies for not thinking to call that out earlier.

iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
jsquire: This has been updated.

README.md
JimacoMS3: @jsquire That sounds like the right thing to do. Thanks!

iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
JimacoMS3: Great. Thanks!

JimacoMS3:
@jsquire Hey Jesse -- I pulled from your fork and tried to run this on my Windows 10 machine. Got the following error:

image

jsquire:
>> I pulled from your fork and tried to run this on my Windows 10 machine. Got the following error:

Thanks for the heads-up.   Looks like the last rebase got a bit twisted and my local copy disagreed with my fork.   Should be all fixed up now.

iot-hub/Quickstarts/read-d2c-messages/README.md
```diff
@@ -0,0 +1,58 @@
...
+The options may be created as follows:
+
+```csharp
+var options = new EventHubConsumerClientOptions();
+
+// This line sets the transport to use WebSockets.
+options.ConnectionOptions.TransportType = EventHubsTransportType.AmqpWebSockets;
+
+// The following lines configure the options for proxy use.
+IWebProxy proxy = { SUPPLY YOUR PROXY };
```
JimacoMS3, line 41: Wouldn't this generally be an instance of [WebProxy](https://docs.microsoft.com/en-us/dotnet/api/system.net.webproxy?view=netcore-3.1)? If so, can we be a little more specific than {SUPPLY YOUR PROXY}?

Also, in addition to the comments in the readme, the samples on the other platforms include some example code for WebSockets and proxy (commented out where necessary). Is it possible to do this for csharp? For example: https://github.com/Azure-Samples/azure-iot-samples-python/pull/24/files or https://github.com/Azure-Samples/azure-iot-samples-java/pull/15/files

jsquire: Fair enough.   Updated.

iot-hub/Quickstarts/read-d2c-messages/ReadDeviceToCloudMessages.cs
```diff
@@ -1,101 +1,111 @@
...
+        // Asynchronously create a PartitionReceiver for a partition and then start
+        // reading any messages sent from the simulated client.
+        private static async Task ReceiveMessagesFromDeviceAsync(CancellationToken cancellationToken)
+        {
+            // If you chose to copy the "Event Hub-compatible endpoint" from the "Built-in endpoints" section
+            // of your IoT Hub instance in the Azure portal, you can set the connection string to that value
+            // directly and remove the call to "BuildEventHubsConnectionString".
+            string connectionString = BuildEventHubsConnectionString(EventHubsCompatibleEndpoint, IotHubSasKeyName, IotHubSasKey);
+
+            // Create the consumer using the default consumer group using a direct connection to the service.
```
jsquire, line 39: There was a comment left at one point, mentioning a desire to include a commented section with proxy use details.   While I'd prefer to avoid that, as I believe that it hurts readability and adds complexity to the sample that aren't needed in the majority case,  I did add a call-out in the comment to point interested parties at the discussion and snippet for proxy support.

jsquire: Note that this link will not work until this PR has merged, as it refers to a new README asset being added in the scope of these changes.

JimacoMS3: Hi Jesse -- we've included the code in Node and Java samples, but you did a great job explaining it in the Readme. I'm OK with not having it in the sample for now. @ramya-rao-a - cool?

README.md
JimacoMS3: > One additional bit of context: this impacts only the recommended SDK for compiling; it does not imply that a new runtime would be needed for deploying. I've left the target framework for the updated sample as `netcoreapp2.1`.
> 
> Apologies for not thinking to call that out earlier.

@jsquire I'm not sure I completely understand here. I thought you meant that only compiling the SDK was impacted, but it seems from the note that you added that in order to compile the samples w/o specifying preview you need 3.0, Is that right? And, if so, it seems like the note would also need to be added in the prereqs in the quickstart as well. Please advise

README.md
```diff
@@ -16,7 +16,9 @@ azure-iot-samples-csharp provides a set of easy-to-understand, continuously-test
...
+
+  **Note:** The samples can be compiled using the NET Core SDK 2.1 SDK if the language version of projects using C# 8 features are changed to `preview`.
```
JimacoMS3, line 21: Please see comment above regarding this note.

README.md
jsquire: Apologies for the confusion.

In order to compile C# code that uses the syntax from v8 (released September 2019), without using the `preview` flag, the .NET Core SDK v3 or higher must be used.   It is possible to use the .NET Core SDK v2.1 to compile if setting the language version to `preview`.  

The Event Hubs client library API was designed around some of the C# 8 concepts, most notably asynchronous enumerators for reading events, which is the scenario covered in the quick start that was updated.  In order to compile the quick start code without modification the .NET Core SDK 3.0+ must be used.   

For developers who wish to use an earlier compiler, they would either need to use the .NET Core SDK 2.1+ and set the language version of the quick start project to `preview` or would need to rewrite the code to manage the enumerable manually, which would look something like the snippet [[here]](https://github.com/Azure/azure-sdk-for-net/issues/9853). 

The distinction that I was trying to make was the difference between which version of the .NET Core SDK was needed to compile the code versus the versions of the .NET Core Runtime or .NET Framework that the output could be used with - meaning which target platforms are supported.  Both the .NET Core SDK v3 and v2.1 can compile such the output can run on [multiple platforms](https://docs.microsoft.com/en-us/dotnet/standard/frameworks) spanning .NET Core, .NET Standard, and the .NET Framework.  

That question is a common concern voiced by developers using the Azure SDKs.   The important take-away here is that the proposed change only impacts folks compiling the quick start (or application) code but does not require that hosting environments be upgraded; any runtime environment that the .NET Core SDK v2.1 supported is also supported by the .NET Core SDK v3 .

jsquire: > And, if so, it seems like the note would also need to be added in the prereqs in the quickstart as well. Please advise

Correct; it appears in the `Prerequisites` section of the README that I added to the quick start.

README.md
jsquire: Likewise, the response for this is included in the above discussion.

jsquire:
:shipit:

ramya-rao-a:
let's go!

JimacoMS3:
#sign-off

@robinsh the associated docs PR is merged and will go live with the 3pm-ish put cycle today (6/2). Try to schedule the merge somewhat close to either side of that time, but we don't have to obsess too much about it. Whatever works for you -- works for me.

PR #121, opened 2020-07-16
fix(samples): Fixing the cleanup script to be case insensitive like A…

vinagesh:
These azure resources are case-insensitive and were getting removed unintentionally by the script if they were created with a different casing.


bikamani:
> These azure resources are case-insensitive and were getting removed unintentionally by the script if they were created with a different casing.

Good to know :)

PR #124, opened 2020-08-07
feat(sample): Add device reconnection sample project

abhipsaMisra:
Add a sample that monitors the device connection status change handler.

Readme doc: https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/abmisr/reconnection/iot-hub/Samples/device/DeviceReconnectionSample/readme.md

### DIFF
--- a/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
@@ -8,46 +8,73 @@ using System.Linq;
 
 namespace Microsoft.Azure.Devices.Client.Samples
 {
+    /// <summary>
+    /// The ILogger implementation for writing color log entries to console.
+    /// </summary>
     public class ColorConsoleLogger : ILogger
     {
         private readonly string _name;
         private readonly ColorConsoleLoggerConfiguration _config;
 
+        /// <summary>
+        /// Initializes an instance of <see cref="ColorConsoleLogger"/>.
+        /// </summary>
+        /// <param name="name">The category name for each <see cref="ILogger"/> instance. This is usually the class name where the logger is initialized.</param>
+        /// <param name="config">The <see cref="ColorConsoleLoggerConfiguration"/> settings to be used for logging.</param>
         public ColorConsoleLogger(string name, ColorConsoleLoggerConfiguration config)
         {
             _name = name;
             _config = config;
         }
 
+        /// <summary>
+        /// Begin a group of logical operations.
+        /// </summary>
+        /// <typeparam name="TState">The type of the state to begin scope for.</typeparam>
+        /// <param name="state">The identifier for the scope.</param>
+        /// <returns></returns>
         public IDisposable BeginScope<TState>(TState state)
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Checks if the given log level is enabled.
+        /// </summary>
+        /// <param name="logLevel">The log level to be checked.</param>
+        /// <returns></returns>
         public bool IsEnabled(LogLevel logLevel)
         {
             return logLevel >= _config.MinLogLevel;
         }
 
+        /// <summary>
+        /// Writes the log entry to console output.
+        /// </summary>
+        /// <typeparam name="TState">The type of the object to be written.</typeparam>
+        /// <param name="logLevel">The log level of the log entry to be written.</param>
+        /// <param name="eventId">The event Id of the log entry to be written.</param>
+        /// <param name="state">The log entry to be written.</param>
+        /// <param name="exception">The exception related to the log entry.</param>
+        /// <param name="formatter">The formatter to be used for formatting the log message.</param>
         public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
         {
-            var colorToBeUsed = _config.LogLevelToColorMapping[logLevel];
-
             if (!IsEnabled(logLevel))
             {
                 return;
             }
 
-            if (_config.EventIds.Contains(0) || _config.EventIds.Contains(eventId.Id))
+            var color = _config.LogLevelToColorMapping[logLevel];
+            if (_config.EventIds.Contains(ColorConsoleLoggerConfiguration.DefaultEventId) || _config.EventIds.Contains(eventId.Id))
             {
-                var color = Console.ForegroundColor;
+                var initialColor = Console.ForegroundColor;
 
                 Console.ForegroundColor = ConsoleColor.Green;
                 Console.Write($"{DateTime.Now.ToString("yyyy-MM-ddTHH:mm:ss.fffffff", CultureInfo.InvariantCulture)}>> ");
 
-                Console.ForegroundColor = colorToBeUsed;
-                Console.WriteLine($"{logLevel} - {formatter(state, exception)}");
                 Console.ForegroundColor = color;
+                Console.WriteLine($"{logLevel} - {formatter(state, exception)}");
+                Console.ForegroundColor = initialColor;
             }
         }
     }
--- a/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
@@ -4,6 +4,7 @@
 using Microsoft.Extensions.Logging;
 using System;
 using System.Globalization;
+using System.Linq;
 
 namespace Microsoft.Azure.Devices.Client.Samples
 {
@@ -25,26 +26,27 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
         public bool IsEnabled(LogLevel logLevel)
         {
-            return logLevel == _config.LogLevel;
+            return logLevel >= _config.MinLogLevel;
         }
 
         public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
         {
+            var colorToBeUsed = _config.LogLevelToColorMapping[logLevel];
+
             if (!IsEnabled(logLevel))
             {
                 return;
             }
 
-            if (_config.EventId == 0 || _config.EventId == eventId.Id)
+            if (_config.EventIds.Contains(0) || _config.EventIds.Contains(eventId.Id))
             {
                 var color = Console.ForegroundColor;
 
                 Console.ForegroundColor = ConsoleColor.Green;
-                Console.Write($"{DateTime.Now.ToString("yyyy-MM-ddTHH:mm:ss.fffffff", CultureInfo.InvariantCulture)}>>");
+                Console.Write($"{DateTime.Now.ToString("yyyy-MM-ddTHH:mm:ss.fffffff", CultureInfo.InvariantCulture)}>> ");
 
-                Console.ForegroundColor = _config.Color;
+                Console.ForegroundColor = colorToBeUsed;
                 Console.WriteLine($"{logLevel} - {formatter(state, exception)}");
-
                 Console.ForegroundColor = color;
             }
         }
--- a/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
@@ -1,0 +1,52 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Extensions.Logging;
+using System;
+using System.Globalization;
+
+namespace Microsoft.Azure.Devices.Client.Samples
+{
+    public class ColorConsoleLogger : ILogger
+    {
+        private readonly string _name;
+        private readonly ColorConsoleLoggerConfiguration _config;
+
+        public ColorConsoleLogger(string name, ColorConsoleLoggerConfiguration config)
+        {
+            _name = name;
+            _config = config;
+        }
+
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return logLevel == _config.LogLevel;
+        }
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            if (!IsEnabled(logLevel))
+            {
+                return;
+            }
+
+            if (_config.EventId == 0 || _config.EventId == eventId.Id)
+            {
+                var color = Console.ForegroundColor;
+
+                Console.ForegroundColor = ConsoleColor.Green;
+                Console.Write($"{DateTime.Now.ToString("yyyy-MM-ddTHH:mm:ss.fffffff", CultureInfo.InvariantCulture)}>>");
+
+                Console.ForegroundColor = _config.Color;
+                Console.WriteLine($"{logLevel} - {formatter(state, exception)}");
+
+                Console.ForegroundColor = color;
+            }
+        }
+    }
+}
--- a/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
@@ -3,20 +3,52 @@
 
 using Microsoft.Extensions.Logging;
 using System;
+using System.Collections.Generic;
 
 namespace Microsoft.Azure.Devices.Client.Samples
 {
     /// <summary>
-    /// A color console logger configuration that sets the default level to Debug and the color to Cyan.
-    /// It also logs all events.
+    /// A color console logger configuration that sets the default log level to Information and logs all events.
     /// </summary>
     public class ColorConsoleLoggerConfiguration
     {
-        // If the EventId is set to 0, we will log all events.
-        private const int EventIdAllEvents = 0;
+        // If the EventId is set to 0, the logger will log all events.
+        private const int DefaultEventId = 0;
 
-        public LogLevel LogLevel { get; set; } = LogLevel.Debug;
-        public int EventId { get; set; } = EventIdAllEvents;
-        public ConsoleColor Color { get; set; } = ConsoleColor.Cyan;
+        private static readonly IDictionary<LogLevel, ConsoleColor> s_defaultColorMapping = new Dictionary<LogLevel, ConsoleColor>
+        {
+            { LogLevel.Trace, ConsoleColor.Blue },
+            { LogLevel.Debug, ConsoleColor.DarkYellow },
+            { LogLevel.Information, ConsoleColor.Cyan },
+            { LogLevel.Warning, ConsoleColor.DarkMagenta },
+            { LogLevel.Error, ConsoleColor.Red },
+            { LogLevel.Critical, ConsoleColor.DarkRed }
+        };
+
+        public ColorConsoleLoggerConfiguration()
+        {
+            LogLevelToColorMapping = s_defaultColorMapping;
+        }
+
+        public ColorConsoleLoggerConfiguration(IDictionary<LogLevel, ConsoleColor> customConsoleColorMapping)
+        {
+            var customMap = s_defaultColorMapping;
+
+            // If a custom color mapping is provided, use it to override the default color mapping.
+            foreach (KeyValuePair<LogLevel, ConsoleColor> entry in customConsoleColorMapping)
+            {
+                if (customMap.ContainsKey(entry.Key))
+                {
+                    customMap[entry.Key] = entry.Value;
+                }
+            }
+            LogLevelToColorMapping = customMap;
+        }
+
+        public IDictionary<LogLevel, ConsoleColor> LogLevelToColorMapping { get; }
+
+        public LogLevel MinLogLevel { get; set; } = LogLevel.Information;
+
+        public IEnumerable<int> EventIds { get; set; } = new List<int>() { DefaultEventId };
     }
 }
--- a/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
@@ -31,18 +31,16 @@ namespace Microsoft.Azure.Devices.Client.Samples
         }
 
         public ColorConsoleLoggerConfiguration(IDictionary<LogLevel, ConsoleColor> customConsoleColorMapping)
+            : this ()
         {
-            var customMap = s_defaultColorMapping;
-
             // If a custom color mapping is provided, use it to override the default color mapping.
             foreach (KeyValuePair<LogLevel, ConsoleColor> entry in customConsoleColorMapping)
             {
-                if (customMap.ContainsKey(entry.Key))
+                if (LogLevelToColorMapping.ContainsKey(entry.Key))
                 {
-                    customMap[entry.Key] = entry.Value;
+                    LogLevelToColorMapping[entry.Key] = entry.Value;
                 }
             }
-            LogLevelToColorMapping = customMap;
         }
 
         public IDictionary<LogLevel, ConsoleColor> LogLevelToColorMapping { get; }
--- a/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
@@ -6,10 +6,17 @@ using System;
 
 namespace Microsoft.Azure.Devices.Client.Samples
 {
+    /// <summary>
+    /// A color console logger configuration that sets the default level to Debug and the color to Cyan.
+    /// It also logs all events.
+    /// </summary>
     public class ColorConsoleLoggerConfiguration
     {
+        // If the EventId is set to 0, we will log all events.
+        private const int EventIdAllEvents = 0;
+
         public LogLevel LogLevel { get; set; } = LogLevel.Debug;
-        public int EventId { get; set; } = 0;
+        public int EventId { get; set; } = EventIdAllEvents;
         public ConsoleColor Color { get; set; } = ConsoleColor.Cyan;
     }
 }
--- a/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
@@ -8,14 +8,14 @@ using System.Collections.Generic;
 namespace Microsoft.Azure.Devices.Client.Samples
 {
     /// <summary>
-    /// A color console logger configuration that sets the default log level to Information and logs all events.
+    /// A color console logger configuration that creates different color console entries per log level, sets the default log level to Information and logs all events.
     /// </summary>
     public class ColorConsoleLoggerConfiguration
     {
         // If the EventId is set to 0, the logger will log all events.
-        private const int DefaultEventId = 0;
+        internal const int DefaultEventId = 0;
 
-        private static readonly IDictionary<LogLevel, ConsoleColor> s_defaultColorMapping = new Dictionary<LogLevel, ConsoleColor>
+        private static readonly IReadOnlyDictionary<LogLevel, ConsoleColor> s_defaultColorMapping = new Dictionary<LogLevel, ConsoleColor>
         {
             { LogLevel.Trace, ConsoleColor.Blue },
             { LogLevel.Debug, ConsoleColor.DarkYellow },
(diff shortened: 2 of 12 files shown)


iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return logLevel == _config.LogLevel;
```
drwill-ms, line 28: I assume logging level is a gradually increasing level (e.g. verbose, info, warning, error), in which case, shouldn't this be >= (or <= depending on the numeric order)?

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            if (!IsEnabled(logLevel))
+            {
+                return;
+            }
+
+            if (_config.EventId == 0 || _config.EventId == eventId.Id)
```
drwill-ms, line 38: What is 0?

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Microsoft.Azure.Devices.Client.Samples
+{
+    public static class ColorConsoleLoggerExtensions
+    {
+        public static ILoggerFactory AddColorConsoleLogger(this ILoggerFactory loggerFactory, ColorConsoleLoggerConfiguration config)
+        {
+            loggerFactory.AddProvider(new ColorConsoleLoggerProvider(config));
+            return loggerFactory;
+        }
+        public static ILoggerFactory AddColorConsoleLogger(this ILoggerFactory loggerFactory)
```
drwill-ms, line 16: newlines between methods

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.Extensions.Logging;
+using System;
+
+namespace Microsoft.Azure.Devices.Client.Samples
+{
+    public static class ColorConsoleLoggerExtensions
+    {
+        public static ILoggerFactory AddColorConsoleLogger(this ILoggerFactory loggerFactory, ColorConsoleLoggerConfiguration config)
+        {
+            loggerFactory.AddProvider(new ColorConsoleLoggerProvider(config));
```
drwill-ms, line 13: What if these are called twice? Do we want two of the same logger added?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -0,0 +1,227 @@
...
+        private static readonly TimeSpan s_sleepDuration = TimeSpan.FromSeconds(5);
+        private static readonly TimeSpan s_operationTimeout = TimeSpan.FromHours(1);
+
+        private readonly string _deviceConnectionString;
+        private readonly TransportType _transportType;
+        private readonly ILogger _logger;
+        private readonly object _lock = new object();
+
+        private static DeviceClient s_deviceClient;
+        private static ConnectionStatus connectionStatus;
```
drwill-ms, line 25: s_?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            _logger.LogDebug($"Connection status changed: status={status}, reason={reason}");
+
+            connectionStatus = status;
+            switch (connectionStatus)
+            {
+                case ConnectionStatus.Connected:
+                    _logger.LogDebug("### The DeviceClient is CONNECTED; all operations will be carried out as normal.");
+
+                    wasConnected = true;
```
drwill-ms, line 47: isConnected?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -0,0 +1,227 @@
...
+                    break;
+
+                case ConnectionStatus.Disconnected_Retrying:
+                    _logger.LogDebug("### The DeviceClient is retrying based on the retry policy. Do NOT close or open the DeviceClient instance");
```
drwill-ms, line 51: shouldn't wasConnected be set to false?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -0,0 +1,227 @@
...
+                case ConnectionStatus.Disabled:
+                    _logger.LogDebug("### The DeviceClient has been closed gracefully." +
+                        "\nIf you want to perform more operations on the device client, you should dispose (DisposeAsync()) and then open (OpenAsync()) the client.");
+
+                    InitializeClient();
+                    break;
+
+                case ConnectionStatus.Disconnected:
+                    {
+                        if (reason == ConnectionStatusChangeReason.Bad_Credential)
```
drwill-ms, line 63: perhaps another switch statement?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -0,0 +1,227 @@
...
+                    await SendMessageAsync(count);
+                }
+                catch (Exception ex) when (ex is IotHubException exception && exception.IsTransient)
+                {
+                    // Inspect the exception to figure out if operation should be retried, or if user-input is required.
+                    _logger.LogError($"An IotHubException was caught, but will try to recover and retry explicitly: {ex}");
+                }
+                catch (Exception ex)
+                {
+                    if (ExceptionHelper.IsNetworkExceptionChain(ex))
```
drwill-ms, line 141: instead use a `when` clause in the exception catch, that way you don't need to throw again on line 147

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -0,0 +1,227 @@
...
+                    {
+                        _logger.LogError($"A network related exception was caught, but will try to recover and retry explicitly: {ex}");
+                    }
+                    else
+                    {
+                        throw;
+                    }
+                }
+
+                await Task.Delay((int)s_sleepDuration.TotalMilliseconds);
```
drwill-ms, line 151: Why not Task.Delay(s_sleepDuration)?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -0,0 +1,227 @@
...
+            }
+        }
+
+        private async Task SendMessageAsync(int count)
+        {
+            var _temperature = s_randomGenerator.Next(20, 35);
+            var _humidity = s_randomGenerator.Next(60, 80);
+            string dataBuffer = $"{{\"temperature\":{_temperature},\"humidity\":{_humidity}}}";
+
+            using var eventMessage = new Message(Encoding.UTF8.GetBytes(dataBuffer))
```
drwill-ms, line 195: Add content encoding and content type properties?

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
drwill-ms: Based on the rest of the code I read, it looks like a console logger is added for a specific level. All this just to have different colors per level?

abhipsaMisra: Yes, all of this is just to add color to the console logs. On directing the sdk logs to the console output it was getting a bit difficult parsing out the info; and the ConsoleLogger available colors only the log level info, and not the log message. 
I'll add a note to the sample highlighting this.

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
abhipsaMisra: EventId when set to 0 will log all events

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerExtensions.cs
abhipsaMisra: so each instance of the console logger here is tagged to a particular log level, so ideally we can have multiple providers added

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: yes, will update

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: Actually this variable is supposed to indicate if the client was ever connected. This is because, if the initial open async fails, then the client reports a Disconnected status, but the underneath layer isn't initialized yet. So when we do a dispose and re-open, it is messing up the internal state, and throws an object disposed exception.
What I can gather from here is, if I call dispose on a client that doesn't have the underlying layer initialized, it is getting messed up.

Now that I think about it more, let me run some tests on the code to see if I can root cause it better. 
This might be a fix only for the symptom, but not the root cause.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: I am setting it inside the block where dispose is being called (after the expression using this variable has been evaluated).

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: sure!

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: good point, yes

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: no reason, I didn't know Task.delay could take in a TimeSpan directtly! I'll use that overload.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: sure, can do

abhipsaMisra:
> For a sample, it could use more documentation about the approach/strategy and why it matters.

Sure, I've added comments around the connection status change handler, and where the exceptions are caught.
Anything specific that I should add, in addition to this?

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
drwill-ms: Why not one logger with colors specified per color? Maybe it could have defaults that are overridable via the ctor?

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
drwill-ms: My point it is a magic number that an uninitiated wouldn't know. Can we make it a constant somewhere discoverable?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: So `wasEverConnected`?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: Probably stemmed from my misunderstanding on what `wasConnected` meant. It should probably be initialized to false and then only changed to true once, right?

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
abhipsaMisra: Sure, that would definitely be an improvement over the current implementation. I can add that in a subsequent PR, since the focus for this one is the reconnection part of the sample. 
Do you think the logging is too confusing, the way it is right now?

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
abhipsaMisra: Sure, I can add it as a constant and add a note explaining what it signifies.

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
drwill-ms: It doesn't work the way I've seen most loggers where logLevel has more to do with reducing the logging output during common activities, so it threw me for a loop. I imagine it would be similarly confusing for others for that reason too.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: sure, sounds better

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: yes, since it is a bool, it defaults to `false`. It is set to true when the client reports `ConnectionStatus.Connected`. After the client is disposed by the calling application, I reset it to false, so that the subsequent initialization and reconnection will again start fresh.

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
abhipsaMisra: I see, I'll see if I can make some quick simplifications, else I'll add the provided ConsoleLogger for now, and add this colored version later.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: makes sense

drwill-ms:
> > For a sample, it could use more documentation about the approach/strategy and why it matters.
> 
> Sure, I've added comments around the connection status change handler, and where the exceptions are caught.
> Anything specific that I should add, in addition to this?

Perhaps a readme that says what the sample is about, why connection management is important, and what are the critical places in code to take note of.

abhipsaMisra:
> > > For a sample, it could use more documentation about the approach/strategy and why it matters.
> > 
> > 
> > Sure, I've added comments around the connection status change handler, and where the exceptions are caught.
> > Anything specific that I should add, in addition to this?
> 
> Perhaps a readme that says what the sample is about, why connection management is important, and what are the critical places in code to take note of.

Sure, I can add a readme!

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            var colorToBeUsed = _config.LogLevelToColorMapping[logLevel];
+
+            if (!IsEnabled(logLevel))
+            {
+                return;
+            }
+
+            if (_config.EventIds.Contains(0) || _config.EventIds.Contains(eventId.Id))
```
drwill-ms, line 41: Still referencing magic number

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
```diff
@@ -0,0 +1,52 @@
...
+            { LogLevel.Critical, ConsoleColor.DarkRed }
+        };
+
+        public ColorConsoleLoggerConfiguration()
+        {
+            LogLevelToColorMapping = s_defaultColorMapping;
+        }
+
+        public ColorConsoleLoggerConfiguration(IDictionary<LogLevel, ConsoleColor> customConsoleColorMapping)
+            : this ()
```
drwill-ms, line 34: Why call the default ctor? It sets a color mapping you don't want.

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
```diff
@@ -0,0 +1,52 @@
...
+            foreach (KeyValuePair<LogLevel, ConsoleColor> entry in customConsoleColorMapping)
+            {
+                if (LogLevelToColorMapping.ContainsKey(entry.Key))
+                {
+                    LogLevelToColorMapping[entry.Key] = entry.Value;
+                }
+            }
+        }
+
+        public IDictionary<LogLevel, ConsoleColor> LogLevelToColorMapping { get; }
```
drwill-ms, line 46: readonly dictionary?

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public LogLevel MinLogLevel { get; set; } = LogLevel.Information;
+
+        public IEnumerable<int> EventIds { get; set; } = new List<int>() { DefaultEventId };
```
drwill-ms, line 50: Why settable?

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
abhipsaMisra: oops, missed this reference, will update

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Extensions.Logging;
+using System;
+using System.Globalization;
+using System.Linq;
+
+namespace Microsoft.Azure.Devices.Client.Samples
+{
+    public class ColorConsoleLogger : ILogger
```
drwill-ms, line 11: Most of these classes lack meaningful comments

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
abhipsaMisra: it sets the default color mapping, and then updates the dic based on what you've passed in-> as in, I might want to override only 1 color mapping, but still retain the default values for others

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -0,0 +1,225 @@
...
+                    _logger.LogDebug("### The DeviceClient is retrying based on the retry policy. Do NOT close or open the DeviceClient instance");
+                    break;
+
+                case ConnectionStatus.Disabled:
+                    _logger.LogDebug("### The DeviceClient has been closed gracefully." +
+                        "\nIf you want to perform more operations on the device client, you should dispose (DisposeAsync()) and then open (OpenAsync()) the client.");
+                    break;
+
+                case ConnectionStatus.Disconnected:
+                    {
```
drwill-ms, line 60: Curious the reason for the extra braces

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
abhipsaMisra: there is a constructor that still updates this dictionalry with any custom mappings supplied

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
abhipsaMisra: I might want to log only specific event Ids, so this is settable but not required. The sample doesn't use this feature, though

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -0,0 +1,225 @@
...
+                    s_deviceClient.OperationTimeoutInMilliseconds = (uint)s_operationTimeout.TotalMilliseconds;
+                    _logger.LogDebug($"Initialized the client instance.");
+                }
+            }
+        }
+
+        private async Task SendMessagesAsync()
+        {
+            int count = 0;
+            while (true)
```
drwill-ms, line 135: So the only exit condition is an unhandled exception?

abhipsaMisra: yes, since I wanted it to be long running (in order to be able to simulate exception and recovery), I did not put a limit on count/ time spent. I had a console input based approach, but while testing it, it was too cumbersome.

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
abhipsaMisra: true, I'll add proper comments

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -0,0 +1,225 @@
...
+                string messageData = Encoding.ASCII.GetString(receivedMessage.GetBytes());
+                _logger.LogInformation($"Received message: {messageData}");
+
+                int propCount = 0;
+                foreach (var prop in receivedMessage.Properties)
+                {
+                    _logger.LogInformation($"Property[{propCount++}> Key={prop.Key} : Value={prop.Value}");
+                }
+
+                await s_deviceClient.CompleteAsync(receivedMessage).ConfigureAwait(false);
```
drwill-ms, line 217: We aren't consistently using ConfigureAwait(false) everywhere.

Do we want to use it in the sample or not?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: yup, doesn't need the extra braces there, will remove

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: We aren't using it in console apps, I'll remove this

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
```diff
@@ -0,0 +1,57 @@
...
+        //private static TransportType s_transportType = TransportType.Mqtt_WebSocket_Only;
+
+        public static async Task<int> Main(string[] args)
+        {
+            // Create a console logger, that logs all events that are categorized at Debug level or higher.
+            // For additional details, see https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger?view=dotnet-plat-ext-3.1.
+            ILoggerFactory loggerFactory = new LoggerFactory();
+
+            loggerFactory.AddColorConsoleLogger(
+                new ColorConsoleLoggerConfiguration()
```
drwill-ms, line 38: ctor parens not required when using an initializer list and ctor has no params

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            if (string.IsNullOrEmpty(s_deviceConnectionString) && args.Length > 0)
+            {
+                s_deviceConnectionString = args[0];
+            }
+
+            var sample = new DeviceReconnectionSample(s_deviceConnectionString, s_transportType, logger);
+            await sample.RunSampleAsync();
+
+            logger.LogInformation("Done, exiting...");
```
drwill-ms, line 53: I'm not clear how this sample gracefully exits, or how this line of code is reached.

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
```diff
@@ -0,0 +1,56 @@
+# Device Reconnection Sample
+
+This sample code demonstrates the various connection status changes and connection status change reasons the Device Client can return, and how to handle them.
```
drwill-ms, line 3: `device client` lower case - not a proper noun

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
```diff
@@ -0,0 +1,56 @@
...
+    <td rowspan="4"> Application </td>
+    <td> This signifies that the device/ module has been deleted or marked as disabled (on your hub instance). <br/> Fix the device/ module status in Azure before attempting to create the associated client instance. </td>
+  </tr>
+  <tr>
+    <td> Bad_Credential </td>
+    <td> Supplied credential isn’t good for device to connect to service. <br/> Fix the supplied credentials before attempting to reconnect again </td>
+  </tr>
+  <tr>
+    <td> Communication_Error </td>
+    <td> This is the state when SDK landed up in a non-retriable error during communication. <br/> If you want to perform more operations on the device client, you should dispose and then re-initialize the client. </td>
```
drwill-ms, line 38: What should be done with pending commands, telemetry that didn't get confirmed as sent, etc.?

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
```diff
@@ -0,0 +1,56 @@
+# Device Reconnection Sample
```
drwill-ms, line 1: This sample shows the connection status code for handling these scenarios, but how does a customer make them happen to see the code in action?

abhipsaMisra: You mean, steps on how to simulate an error, so that the client does retry/ recover? I'll add some notes on that, sure.

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
```diff
@@ -0,0 +1,56 @@
...
+
+This sample code demonstrates the various connection status changes and connection status change reasons the Device Client can return, and how to handle them.
+
+The device client exhibits the following connection status changes with reason:
```
drwill-ms, line 5: Essentially, connection status changes come down to a few actions, right?

1. Use the client
2. Wait to use the client
3. Reinitialize the client
4. Abandon the app - requires human intervention

Right?

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
abhipsaMisra: The call wouldn't complete, so technically the only pending item is the operation that was being executed when the error happened. I'll add a note explaining that, sure.

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
abhipsaMisra: It doesn't, right now. I'll update the sample to run for a pre-set amount of time.

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
abhipsaMisra: yes, correct

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
drwill-ms: got it.

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
drwill-ms: You'd need to cast the property then.

The question is, do you want the user to be able to update the dictionary after init?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: How about catching control-c in the console, and signaling a CancellationToken?

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
drwill-ms: Presumably there could be lots of other pending actions right?

I run into a communication error at time T0, and my retries are exhausted at time T10. From T0 to T9 I had lots of initiated telemetry events and commands received. They are all pending due to the comm error that started at T0.

If I dispose my client and init, I've lost all that stuff right?

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
drwill-ms: Yep, like cycle a device primary key, pull network cable, etc.

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
drwill-ms: I'd make that clear in this readme, and for each connection status reference this enum.

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
abhipsaMisra: No, I don't. I want this dict to be initialized only in the constructor.

abhipsaMisra: would that then be something like this?
```csharp
public ColorConsoleLoggerConfiguration(IDictionary<LogLevel, ConsoleColor> customConsoleColorMapping)
	: this ()
{
	var map = (IDictionary<LogLevel, ConsoleColor> )LogLevelToColorMapping;

	// If a custom color mapping is provided, use it to override the default color mapping.
	foreach (KeyValuePair<LogLevel, ConsoleColor> entry in customConsoleColorMapping)
	{
		if (map.ContainsKey(entry.Key))
		{
			map[entry.Key] = entry.Value;
		}
	}
	LogLevelToColorMapping = (IReadOnlyDictionary<LogLevel, ConsoleColor>)map;
}
```

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: sure, I prefer this to setting a pre-determined timeout, I'll look into that

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
abhipsaMisra: will be addressed as a part of https://github.com/Azure-Samples/azure-iot-samples-csharp/pull/124#discussion_r468729862

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
abhipsaMisra: Are we assuming that these telemetry operations were initiated by separate tasks, or by the same task? 
I'll sync up with you offline on this, I might be misinterpreting your question.

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
drwill-ms: You could avoid the first line (casting the existing, initialized) RO dict to an IDict by putting all the initialized values in the ctor, but what you also have here is fine.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: https://github.com/Azure/azure-digital-twins/blob/private-preview/Demo/DeviceSimulator/DeviceSimulator/Program.cs

PR #128, opened 2020-08-28
Fix for import devices sample

drwill-ms:
Customer issue was logged about this sample not working for them, yielding an error from the Hub of `BlobContainerValidationError`.

I noticed the container was empty when I expected there to be a file there, so I found a Commit method on the stream. After calling that, the file is written and the job succeeds.

I included some misc other code cleanup.

### DIFF
--- a/iot-hub/Samples/service/ImportExportDevicesSample/ImportExportDevicesSample.cs
+++ b/iot-hub/Samples/service/ImportExportDevicesSample/ImportExportDevicesSample.cs
@@ -16,7 +16,6 @@ namespace Microsoft.Azure.Devices.Samples
 {
     public class ImportExportDevicesSample
     {
-
         // Container used to hold the blob containing the list of import/export files.
         // This is a module-wide variable. If this project doesn't find this container, it will create it.
         private CloudBlobContainer _cloudBlobContainer;
@@ -34,7 +33,10 @@ namespace Microsoft.Azure.Devices.Samples
         private readonly string _destIotHubConnectionString;
         private readonly string _storageAccountConnectionString;
 
-        public ImportExportDevicesSample(string sourceIotHubConnectionString, string destinationHubConnectionString, string sourceStorageAccountConnectionString)
+        public ImportExportDevicesSample(
+            string sourceIotHubConnectionString,
+            string destinationHubConnectionString,
+            string sourceStorageAccountConnectionString)
         {
             _destIotHubConnectionString = destinationHubConnectionString;
             _iotHubConnectionString = sourceIotHubConnectionString;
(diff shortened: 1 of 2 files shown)


iot-hub/Samples/service/ImportExportDevicesSample/ImportExportDevicesSample.cs
```diff
@@ -137,20 +138,21 @@ private async Task<string> PrepareStorageForImportExportAsync(string storageAcco
...
-            sasConstraints.Permissions = SharedAccessBlobPermissions.Write
-                | SharedAccessBlobPermissions.Read
-                | SharedAccessBlobPermissions.Delete;
+            var sasConstraints = new SharedAccessBlobPolicy
+            {
+                SharedAccessExpiryTime = DateTime.UtcNow.AddHours(24),
+                Permissions = SharedAccessBlobPermissions.Write
+                    | SharedAccessBlobPermissions.Read
+                    | SharedAccessBlobPermissions.Delete
+                    | SharedAccessBlobPermissions.Create,
```
drwill-ms, line 154: In my testing, I added create. I probably should remove it, as it doesn't seem to be necessary.

iot-hub/Samples/service/ImportExportDevicesSample/ImportExportDevicesSample.cs
```diff
@@ -168,11 +170,9 @@ private string GetContainerSasToken(CloudBlobContainer container)
         ///    then use this the Copy feature to copy the devices to another hub.
         /// Number of devices to create and add. Default is 10.
         /// </summary>
-        /// <returns></returns>
         public async Task AddNewDevices(string hubConnectionString, int NumToAdd)
         {
-            Stopwatch stopwatch = new Stopwatch();
-            stopwatch.Start();
+            Stopwatch stopwatch = Stopwatch.StartNew();
```
drwill-ms, line 175: Why code in 2 lines, when 1 line will do? :)

iot-hub/Samples/service/ImportExportDevicesSample/ImportExportDevicesSample.cs
```diff
@@ -308,37 +303,37 @@ private async Task GenerateAndAddDevices(string hubConnectionString, string cont
...
-                byte[] bytes = Encoding.UTF8.GetBytes(sb.ToString());
-                for (var i = 0; i < bytes.Length; i += 500)
-                {
-                    int length = Math.Min(bytes.Length - i, 500);
-                    await stream.WriteAsync(bytes, i, length).ConfigureAwait(false);
-                }
+                int length = Math.Min(bytes.Length - i, 500);
+                await stream.WriteAsync(bytes, i, length).ConfigureAwait(false);
             }
+            stream.Commit();
```
drwill-ms, line 313: This was the key that made the sample start working. I found other places, like here, to add it as well.

PR #129, opened 2020-09-04
Hardcode client version, improve TransportType selection, remove Conf…

drwill-ms:
Hardcode client version, improve TransportType selection, remove ConfigureAwait, and misc code cleanup

### DIFF
--- a/iot-hub/Samples/common/DeviceStreamingCommon.cs
+++ b/iot-hub/Samples/common/DeviceStreamingCommon.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Devices.Samples.Common
             var wsClient = new ClientWebSocket();
             wsClient.Options.SetRequestHeader("Authorization", "Bearer " + authorizationToken);
 
-            await wsClient.ConnectAsync(url, cancellationToken).ConfigureAwait(false);
+            await wsClient.ConnectAsync(url, cancellationToken);
 
             return wsClient;
         }
--- a/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
@@ -13,7 +13,6 @@ namespace Microsoft.Azure.Devices.Client.Samples
     /// </summary>
     public class ColorConsoleLogger : ILogger
     {
-        private readonly string _name;
         private readonly ColorConsoleLoggerConfiguration _config;
 
         /// <summary>
@@ -21,9 +20,8 @@ namespace Microsoft.Azure.Devices.Client.Samples
         /// </summary>
         /// <param name="name">The category name for each <see cref="ILogger"/> instance. This is usually the class name where the logger is initialized.</param>
         /// <param name="config">The <see cref="ColorConsoleLoggerConfiguration"/> settings to be used for logging.</param>
-        public ColorConsoleLogger(string name, ColorConsoleLoggerConfiguration config)
+        public ColorConsoleLogger(ColorConsoleLoggerConfiguration config)
         {
-            _name = name;
             _config = config;
         }
 
@@ -32,7 +30,6 @@ namespace Microsoft.Azure.Devices.Client.Samples
         /// </summary>
         /// <typeparam name="TState">The type of the state to begin scope for.</typeparam>
         /// <param name="state">The identifier for the scope.</param>
-        /// <returns></returns>
         public IDisposable BeginScope<TState>(TState state)
         {
             throw new NotImplementedException();
@@ -42,7 +39,6 @@ namespace Microsoft.Azure.Devices.Client.Samples
         /// Checks if the given log level is enabled.
         /// </summary>
         /// <param name="logLevel">The log level to be checked.</param>
-        /// <returns></returns>
         public bool IsEnabled(LogLevel logLevel)
         {
             return logLevel >= _config.MinLogLevel;
--- a/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
@@ -15,48 +15,18 @@ namespace Microsoft.Azure.Devices.Client.Samples
         // If the EventId is set to 0, the logger will log all events.
         internal const int DefaultEventId = 0;
 
-        private static readonly IReadOnlyDictionary<LogLevel, ConsoleColor> s_defaultColorMapping = new Dictionary<LogLevel, ConsoleColor>
+        /// <summary>
+        /// A dictionary containing the log level to console color mappings to be used while writing log entries to the console.
+        /// </summary>
+        public IReadOnlyDictionary<LogLevel, ConsoleColor> LogLevelToColorMapping { get; } = new Dictionary<LogLevel, ConsoleColor>
         {
             { LogLevel.Trace, ConsoleColor.Blue },
             { LogLevel.Debug, ConsoleColor.DarkYellow },
             { LogLevel.Information, ConsoleColor.Cyan },
             { LogLevel.Warning, ConsoleColor.DarkMagenta },
             { LogLevel.Error, ConsoleColor.Red },
-            { LogLevel.Critical, ConsoleColor.DarkRed }
+            { LogLevel.Critical, ConsoleColor.DarkRed },
         };
-
-        /// <summary>
-        /// Initialize an instance of <see cref="ColorConsoleLoggerConfiguration"/> with default color mappings.
-        /// </summary>
-        public ColorConsoleLoggerConfiguration()
-        {
-            LogLevelToColorMapping = s_defaultColorMapping;
-        }
-
-        /// <summary>
-        /// Initialize an instance of <see cref="ColorConsoleLoggerConfiguration"/> by overriding the default color mappings with the supplied custom mappings.
-        /// </summary>
-        /// <param name="customConsoleColorMapping">A dictionary of log level to console color mapping that will be used to override the default color mapping.</param>
-        public ColorConsoleLoggerConfiguration(IDictionary<LogLevel, ConsoleColor> customConsoleColorMapping)
-            : this ()
-        {
-            var map = (IDictionary<LogLevel, ConsoleColor> )LogLevelToColorMapping;
-
-            // If a custom color mapping is provided, use it to override the default color mapping.
-            foreach (KeyValuePair<LogLevel, ConsoleColor> entry in customConsoleColorMapping)
-            {
-                if (map.ContainsKey(entry.Key))
-                {
-                    map[entry.Key] = entry.Value;
-                }
-            }
-            LogLevelToColorMapping = (IReadOnlyDictionary<LogLevel, ConsoleColor>)map;
-        }
-
-        /// <summary>
-        /// A dictionary containing the log level to console color mappings to be used while writing log entries to the console.
-        /// </summary>
-        public IReadOnlyDictionary<LogLevel, ConsoleColor> LogLevelToColorMapping { get; }
 
         /// <summary>
         /// The min log level that will be written to the console, defaults to <see cref="LogLevel.Information"/>.
@@ -66,6 +36,6 @@ namespace Microsoft.Azure.Devices.Client.Samples
         /// <summary>
         /// The list of event Ids to be written to the console. By default, all event Ids are written.
         /// </summary>
-        public IEnumerable<int> EventIds { get; set; } = new List<int>() { DefaultEventId };
+        public IEnumerable<int> EventIds { get; } = new int[] { DefaultEventId };
     }
 }
--- a/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
@@ -35,6 +35,52 @@ namespace Microsoft.Azure.Devices.Client.Samples
             InitializeClient();
         }
 
+        public async Task RunSampleAsync()
+        {
+            using var cts = new CancellationTokenSource();
+
+            Console.CancelKeyPress += (sender, eventArgs) =>
+            {
+                eventArgs.Cancel = true;
+                cts.Cancel();
+                _logger.LogInformation("Sample execution cancellation requested; will exit.");
+            };
+
+            try
+            {
+                await Task.WhenAll(SendMessagesAsync(cts.Token), ReceiveMessagesAsync(cts.Token));
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError($"Unrecoverable exception caught, user action is required, so exiting...: \n{ex}");
+            }
+        }
+
+        private void InitializeClient()
+        {
+            // If the client reports Connected status, it is already in operational state.
+            if (s_connectionStatus != ConnectionStatus.Connected)
+            {
+                lock (_lock)
+                {
+                    _logger.LogDebug($"Attempting to initialize the client instance, current status={s_connectionStatus}");
+
+                    // If the device client instance has been previously initialized, then dispose it.
+                    // The s_wasEverConnected variable is required to store if the client ever reported Connected status.
+                    if (s_wasEverConnected && s_connectionStatus == ConnectionStatus.Disconnected)
+                    {
+                        s_deviceClient?.Dispose();
+                        s_wasEverConnected = false;
+                    }
+
+                    s_deviceClient = DeviceClient.CreateFromConnectionString(_deviceConnectionString, _transportType);
+                    s_deviceClient.SetConnectionStatusChangesHandler(ConnectionStatusChangeHandler);
+                    s_deviceClient.OperationTimeoutInMilliseconds = (uint)s_operationTimeout.TotalMilliseconds;
+                    _logger.LogDebug($"Initialized the client instance.");
+                }
+            }
+        }
+
         private void ConnectionStatusChangeHandler(ConnectionStatus status, ConnectionStatusChangeReason reason)
         {
             _logger.LogDebug($"Connection status changed: status={status}, reason={reason}");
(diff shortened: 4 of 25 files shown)


iot-hub/Samples/device/FileUploadSample/FileUploadSample.cs
```diff
@@ -21,20 +19,24 @@ public FileUploadSample(DeviceClient deviceClient)
...
-                watch.Stop();
+            var fileUploadTime = Stopwatch.StartNew();
 
-                Console.WriteLine("Time to upload file: {0}ms\n", watch.ElapsedMilliseconds);
-            }
+            // TODO: update this sample to use the new APIs
+
+#pragma warning disable CS0618 // Type or member is obsolete
+            // Note: UploadToBlobAsync will use HTTPS as protocol, regardless of the DeviceClient protocol selection.
+            await _deviceClient.UploadToBlobAsync(fileName, fileStreamSource);
```
drwill-ms, line 35: This spot will be for @timtay-microsoft

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
```diff
@@ -22,11 +22,15 @@ public class Program
...
+        //   Http1
+        //   Amqp_WebSocket_Only
+        //   Amqp_Tcp_only
+        //   Mqtt
+        //   Mqtt_WebSocket_Only
+        //   Mqtt_Tcp_Only
+        private static readonly string s_transportType = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_TRANSPORT_TYPE");
 
         public static async Task<int> Main(string[] args)
         {
```
drwill-ms, line 36: @abhipsaMisra on line 47 we instantiate a ConsoleEventListener, but never use it. I didn't want to remove it until I asked you what the point was.

iot-hub/Samples/common/DeviceStreamingCommon.cs
```diff
@@ -22,7 +22,7 @@ public static async Task<ClientWebSocket> GetStreamingClientAsync(Uri url, strin
             var wsClient = new ClientWebSocket();
             wsClient.Options.SetRequestHeader("Authorization", "Bearer " + authorizationToken);
 
-            await wsClient.ConnectAsync(url, cancellationToken).ConfigureAwait(false);
+            await wsClient.ConnectAsync(url, cancellationToken);
```
drwill-ms, line 25: For samples, we'll leave off the ConfigureAwait(false) call, to match the advice we got in the mono repo.

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLogger.cs
```diff
@@ -13,17 +13,15 @@ namespace Microsoft.Azure.Devices.Client.Samples
...
     {
-        private readonly string _name;
         private readonly ColorConsoleLoggerConfiguration _config;
 
         /// <summary>
         /// Initializes an instance of <see cref="ColorConsoleLogger"/>.
         /// </summary>
         /// <param name="name">The category name for each <see cref="ILogger"/> instance. This is usually the class name where the logger is initialized.</param>
         /// <param name="config">The <see cref="ColorConsoleLoggerConfiguration"/> settings to be used for logging.</param>
-        public ColorConsoleLogger(string name, ColorConsoleLoggerConfiguration config)
```
drwill-ms, line 24: `name` was unused

iot-hub/Samples/device/DeviceReconnectionSample/ColorConsoleLoggerConfiguration.cs
```diff
@@ -15,48 +15,45 @@ public class ColorConsoleLoggerConfiguration
...
             { LogLevel.Trace, ConsoleColor.Blue },
             { LogLevel.Debug, ConsoleColor.DarkYellow },
             { LogLevel.Information, ConsoleColor.Cyan },
             { LogLevel.Warning, ConsoleColor.DarkMagenta },
             { LogLevel.Error, ConsoleColor.Red },
-            { LogLevel.Critical, ConsoleColor.DarkRed }
+            { LogLevel.Critical, ConsoleColor.DarkRed },
         };
 
+        private Dictionary<LogLevel, ConsoleColor> _logLevelToColorMapping = s_defaultColorMapping;
```
drwill-ms, line 28: Doing this prevents needing to cast in the ctor

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
```diff
@@ -47,10 +51,22 @@ public static async Task<int> Main(string[] args)
                 s_deviceConnectionString = args[0];
             }
 
-            var sample = new DeviceReconnectionSample(s_deviceConnectionString, s_transportType, logger);
+            // Check environment variable for transport type,
+            if (!Enum.TryParse(s_transportType, out TransportType transportType)
```
drwill-ms, line 55: Elevates transport to an environment variable/command-line var.

We really should update these samples to use a command-line library we like, however.

iot-hub/Samples/device/DeviceStreamingSample/DeviceStreamSample.cs
```diff
@@ -21,48 +21,44 @@ public DeviceStreamSample(DeviceClient deviceClient)
...
-                    else
-                    {
-                        await _deviceClient.RejectDeviceStreamRequestAsync(streamRequest, cancellationTokenSource.Token).ConfigureAwait(false);
-                    }
+                    await webSocket
+                        .CloseAsync(WebSocketCloseStatus.NormalClosure, string.Empty, cts.Token);
                 }
 
-                await _deviceClient.CloseAsync().ConfigureAwait(false);
+                await _deviceClient.CloseAsync();
```
drwill-ms, line 61: @abhipsaMisra you think we should not call close on our clients in the samples?

iot-hub/Samples/device/KeysRolloverSample/KeyRolloverSample.cs
```diff
@@ -27,35 +27,36 @@ public async Task RunSampleAsync()
...
-                using (var testMessage = new Message(Encoding.UTF8.GetBytes("message from key rollover sample")))
-                {
-                    for (int i = 0; i * delaySeconds < loopSeconds; i++)
-                    {
-                        Console.WriteLine($"\t {DateTime.Now.ToLocalTime()} Sending message [{usedConnectionString} connection string].");
-                        await deviceClient.SendEventAsync(testMessage).ConfigureAwait(false);
-                        await Task.Delay(delaySeconds * 1000);
-                    }
-                }
+                ContentType = "text/plain",
```
drwill-ms, line 51: We should always specify content type and encoding in our samples

iot-hub/Samples/device/TwinSample/TwinSample.cs
```diff
@@ -18,19 +18,19 @@ public TwinSample(DeviceClient deviceClient)
...
-            Twin twin = await _deviceClient.GetTwinAsync().ConfigureAwait(false);
+            Twin twin = await _deviceClient.GetTwinAsync();
 
             Console.WriteLine("\tInitial twin value received:");
             Console.WriteLine($"\t{twin.ToJson()}");
 
             Console.WriteLine("Sending sample start time as reported property");
             TwinCollection reportedProperties = new TwinCollection();
-            reportedProperties["DateTimeLastAppLaunch"] = DateTime.Now;
+            reportedProperties["DateTimeLastAppLaunch"] = DateTime.UtcNow;
```
drwill-ms, line 31: Unless presenting dates to a user, they should be saved in UTC, especially when written to a global service

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
timtay-microsoft: Is this case sensitive?

iot-hub/Samples/device/TwinSample/TwinSample.csproj
```diff
@@ -3,10 +3,11 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.1</TargetFramework>
+    <LangVersion>8.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
-	<PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.*" />
+	<PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.28.0" />
```
timtay-microsoft, line 10: Why aren't we leaving this as "1.*" to always get the latest?

timtay-microsoft (approved):

azabbasi (approved):

iot-hub/Samples/device/FileUploadSample/FileUploadSample.cs
timtay-microsoft:
```csharp
        public async Task RunSampleAsync()
        {
            const string filePath = "TestPayload.txt";

            using var fileStreamSource = new FileStream(filePath, FileMode.Open);
            var fileName = Path.GetFileName(fileStreamSource.Name);

            var watch = Stopwatch.StartNew();

            var fileUploadSasUriRequest = new FileUploadSasUriRequest()
            {
                BlobName = fileName
            };

            // Note: GetFileUploadSasUriAsync and CompleteFileUploadAsync will use HTTPS as protocol regardless of the DeviceClient protocol selection.
            Console.WriteLine("Getting SAS URI from IoT Hub to use when uploading the file...");
            var fileUploadSasUriResponse = await _deviceClient.GetFileUploadSasUriAsync(fileUploadSasUriRequest);

            Uri sasUri = new Uri(String.Format(
                CultureInfo.InvariantCulture,
                "https://{0}/{1}/{2}{3}",
                fileUploadSasUriResponse.HostName,
                fileUploadSasUriResponse.ContainerName,
                Uri.EscapeDataString(fileUploadSasUriResponse.BlobName), // Pass URL encoded device name and blob name to support special characters
                fileUploadSasUriResponse.SasToken));

            Console.WriteLine("Successfully got SAS URI from IoT Hub: {0}", sasUri);

            try 
            {
                Console.WriteLine("Uploading File: {0} using the Azure Storage SDK and the retrieved SAS URI for authentication", fileName);

                // Note that other versions of the Azure Storage SDK can be used here. For the latest version, see
                // https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/storage#azure-storage-libraries-for-net
                CloudBlockBlob blob = new CloudBlockBlob(sasUri);
                var uploadTask = blob.UploadFromStreamAsync(fileStreamSource);
            }
            catch (Exception e) 
            {
                Console.WriteLine("Failed to upload file to Azure Storage using the Azure Storage SDK: {0}", e);

                var failedFileUploadCompletionNotification = new FileUploadCompletionNotification()
                {
                    // Mandatory. Must be the same value as the correlation id returned in the sas uri response
                    CorrelationId = fileUploadSasUriResponse.CorrelationId,

                    // Mandatory. Will be present when service client receives this file upload notification
                    IsSuccess = false,

                    // Optional, user defined status code. Will be present when service client receives this file upload notification
                    StatusCode = 500,

                    // Optional, user defined status description. Will be present when service client receives this file upload notification
                    StatusDescription = e.Message
                };

                // Note that this is done even when the file upload fails. IoT Hub has a fixed number of SAS URIs allowed active
                // at any given time. Once you are done with the file upload, you should free your SAS URI so that other
                // SAS URIs can be generated. If a SAS URI is not freed through this API, then it will free itself eventually
                // based on how long SAS URIs are configured to live on your IoT Hub.
                Console.WriteLine("Notifying IoT Hub that the file upload failed and that the SAS URI can be freed");
                await _deviceClient.CompleteFileUploadAsync(failedFileUploadCompletionNotification);
                Console.WriteLine("Notified IoT Hub that the file upload failed and that the SAS URI can be freed successfully");

                watch.Stop();
                return;
            }

            Console.WriteLine("Successfully uploaded the file to Azure Storage");

            var successfulFileUploadCompletionNotification = new FileUploadCompletionNotification()
            {
                // Mandatory. Must be the same value as the correlation id returned in the sas uri response
                CorrelationId = fileUploadSasUriResponse.CorrelationId,

                // Mandatory. Will be present when service client receives this file upload notification
                IsSuccess = true,

                // Optional, user defined status code. Will be present when service client receives this file upload notification
                StatusCode = 200,

                // Optional, user defined status description. Will be present when service client receives this file upload notification
                StatusDescription = "File upload was a success!"
            };

            Console.WriteLine("Notifying IoT Hub that the file upload succeeded and that the SAS URI can be freed");
            await _deviceClient.CompleteFileUploadAsync(successfulFileUploadCompletionNotification);
            Console.WriteLine("Notified IoT Hub that the file upload succeeded and that the SAS URI can be freed successfully");

            watch.Stop();

            Console.WriteLine("Time to upload file: {0}ms\n", watch.ElapsedMilliseconds);
        }
```

drwill-ms: @timtay-microsoft we're using the oldest version of the storage SDK here. Can we find one done in this millennium?

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
drwill-ms: Yes

iot-hub/Samples/device/TwinSample/TwinSample.csproj
drwill-ms: Because in .NET that is not considered a best practice. The samples could possibly break without us knowing. We should be explicit about our package updates.

drwill-ms: Also, we don't want to imply that a customer should do that in their project, either. Again, being a bad practice in .NET.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -147,10 +147,10 @@ private void InitializeClient()
         private async Task SendMessagesAsync(CancellationToken cancellationToken)
         {
             int count = 0;
+
             while (!cancellationToken.IsCancellationRequested)
             {
-                count++;
-                _logger.LogInformation($"Device sending message {count} to IoTHub...");
+                _logger.LogInformation($"Device sending message {count++} to IoTHub...");
```
abhipsaMisra, line 153: +1

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
```diff
@@ -22,11 +22,15 @@ public class Program
         // - create a launchSettings.json (see launchSettings.json.template) containing the variable
         private static string s_deviceConnectionString = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_CONN_STRING");
 
-        // Select one of the following transports used by DeviceClient to connect to IoT Hub.
-        private static TransportType s_transportType = TransportType.Amqp;
-        //private static TransportType s_transportType = TransportType.Mqtt;
-        //private static TransportType s_transportType = TransportType.Amqp_WebSocket_Only;
-        //private static TransportType s_transportType = TransportType.Mqtt_WebSocket_Only;
+        // Specify one of the following transports used by DeviceClient to connect to IoT Hub.
+        //   Amqp
```
abhipsaMisra, line 26: nit: why not group the amqp options together?

iot-hub/Samples/device/DeviceStreamingSample/DeviceStreamSample.cs
abhipsaMisra: We usually try to initialize the device client inside a disposable block, and pass the client to the sample program.  If that is not possible, I would recommend that our sample `.dispose()` the client as a part of cleanup, rather than `.close()`.

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
drwill-ms: I put them in the order they are in the enum.

iot-hub/Samples/device/DeviceStreamingSample/DeviceStreamSample.cs
drwill-ms: Dispose is called from Program.cs via the `using` keyword.

So my question is, should I remove this call?

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
drwill-ms: It isn't now.

iot-hub/Samples/device/DeviceReconnectionSample/Program.cs
abhipsaMisra: For the console event listener, we only need to initialize an instance of that class. Since we define custom implementations of the write events from `EventListener`, the appropriate action is taken whenever our library code writes to event source. I cannot say that I am super confident in how EventSource works, but this has been my finding.

abhipsaMisra (approved):

iot-hub/Samples/device/FileUploadSample/FileUploadSample.cs
```diff
@@ -21,20 +21,92 @@ public FileUploadSample(DeviceClient deviceClient)
...
+            var fileUploadTime = Stopwatch.StartNew();
+            
+            var fileUploadSasUriRequest = new FileUploadSasUriRequest
+            {
+                BlobName = fileName
+            };
+
+            // Note: GetFileUploadSasUriAsync and CompleteFileUploadAsync will use HTTPS as protocol regardless of the DeviceClient protocol selection.
+            Console.WriteLine("Getting SAS URI from IoT Hub to use when uploading the file...");
+            var sasUri = await _deviceClient.GetFileUploadSasUriAsync(fileUploadSasUriRequest);
```
drwill-ms, line 40: The sample fails here @timtay-microsoft. Any idea?

> '{"Message":"{\"errorCode\":400022,\"trackingId\":\"315c19c7c6fb495a9786e7dd0ae935c4-G:2-TimeStamp:09/05/2020 00:56:11\",\"message\":\"BadRequest\",\"timestampUtc\":\"2020-09-05T00:56:11.0393084Z\"}","ExceptionMessage":""}'

> Microsoft.Azure.Devices.Client.Transport.HttpClientHelper.ExecuteAsync(System.Net.Http.HttpMethod, System.Uri, System.Func<System.Net.Http.HttpRequestMessage, System.Threading.CancellationToken, System.Threading.Tasks.Task>, System.Func<System.Net.Http.HttpResponseMessage, bool>, System.Func<System.Net.Http.HttpResponseMessage, System.Threading.CancellationToken, System.Threading.Tasks.Task>, System.Collections.Generic.IDictionary<System.Net.HttpStatusCode, System.Func<System.Net.Http.HttpResponseMessage, System.Threading.Tasks.Task<System.Exception>>>, System.Threading.CancellationToken)

timtay-microsoft: I'm looking into this.

drwill-ms: I didn't have my hub configured properly. I fixed that and added instructions in program.cs.

Next failure happens when completing the file upload.

drwill-ms: > System.ArgumentException: '{"Message":"{\"errorCode\":400020,\"trackingId\":\"defa6fd07c2a48deb0ce6398fa452065-G:6-TimeStamp:09/08/2020 19:18:16\",\"message\":\"BadRequest\",\"timestampUtc\":\"2020-09-08T19:18:16.9608672Z\"}","ExceptionMessage":""}'

```
This exception was originally thrown at this call stack:
    Microsoft.Azure.Devices.Client.Transport.HttpClientHelper.ExecuteAsync(System.Net.Http.HttpMethod, System.Uri, System.Func<System.Net.Http.HttpRequestMessage, System.Threading.CancellationToken, System.Threading.Tasks.Task>, System.Func<System.Net.Http.HttpResponseMessage, bool>, System.Func<System.Net.Http.HttpResponseMessage, System.Threading.CancellationToken, System.Threading.Tasks.Task>, System.Collections.Generic.IDictionary<System.Net.HttpStatusCode, System.Func<System.Net.Http.HttpResponseMessage, System.Threading.Tasks.Task<System.Exception>>>, System.Threading.CancellationToken)
    System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
    System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
    Microsoft.Azure.Devices.Client.Transport.HttpTransportHandler.CompleteFileUpload(Microsoft.Azure.Devices.Client.Transport.FileUploadCompletionNotification, System.Threading.CancellationToken)
    System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
    System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
    Microsoft.Azure.Devices.Client.InternalClient.CompleteFileUploadAsync(Microsoft.Azure.Devices.Client.Transport.FileUploadCompletionNotification, System.Threading.CancellationToken)
    System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
    System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
    Microsoft.Azure.Devices.Client.Samples.FileUploadSample.RunSampleAsync() in FileUploadSample.cs
    ...
    [Call Stack Truncated]
```

timtay-microsoft: Seems like an issue with the file that is being uploaded. When I run the sample with a full directory path to a different file (outside of the sample project), everything works as expected. Maybe just make the file path a parameter like the device connection string?

timtay-microsoft: I have no clue why only the last service call would throw an exception because of this, though

PR #133, opened 2020-09-11
Fix the 'threading' code.

robinsh:
The code changes make the multithreading in the routing tutorial match the arm-read-write multithreading (which Robert Outlaw helped me fix).


horizontal-arm/dotnet
```diff
@@ -0,0 +1,4 @@
+Microsoft Windows [Version 10.0.19041.508]
+(c) 2020 Microsoft Corporation. All rights reserved.
+
+C:\_github\azure-iot-samples-csharp\horizontal-arm>
```
drwill-ms, line 4: Did you intend to add this?

iot-hub/Tutorials/Routing/SimulatedDevice/Program.cs
```diff
@@ -60,23 +56,30 @@ private static void Main(string[] args)
                 //  http://docs.microsoft.com/azure/iot-hub/tutorial-routing
 
                 Console.WriteLine("Routing Tutorial: Simulated device\n");
-                s_deviceClient = DeviceClient.Create(s_iotHubUri, new DeviceAuthenticationWithRegistrySymmetricKey(s_myDeviceId, s_deviceKey), TransportType.Mqtt);
-                SendDeviceToCloudMessagesAsync();
+                s_deviceClient = DeviceClient.Create(s_iotHubUri, 
+                  new DeviceAuthenticationWithRegistrySymmetricKey(s_myDeviceId, s_deviceKey), TransportType.Mqtt);
```
drwill-ms, line 60: you don't want to be checking in hub URIs and device keys.

Try something like [this](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/DeviceReconnectionSample/Program.cs#L24).

iot-hub/Tutorials/Routing/SimulatedDevice/Program.cs
```diff
@@ -60,23 +56,30 @@ private static void Main(string[] args)
...
+
+                var cts = new CancellationTokenSource();
```
drwill-ms, line 62: using (requires disposing)

iot-hub/Tutorials/Routing/SimulatedDevice/Program.cs
```diff
@@ -60,23 +56,30 @@ private static void Main(string[] args)
...
         {
             double minTemperature = 20;
             double minHumidity = 60;
             Random rand = new Random();
 
-            while (true)
+            while (!token.IsCancellationRequested)
             {
                 double currentTemperature = minTemperature + rand.NextDouble() * 15;
                 double currentHumidity = minHumidity + rand.NextDouble() * 20;
```
drwill-ms, line 85: line 128, Message requires a using because it requires disposing

iot-hub/Tutorials/Routing/SimulatedDevice/SimulatedDevice.csproj
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.*" />
+    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.*.0" />
```
drwill-ms, line 9: I've been changing our samples out from using this kind of version notation. We should explicitly update them, rather than take the chance a customer picks up the sample and has it fail because it gets one we haven't tested with. Sure, we shouldn't have breaking changes, but I don't want to stake a positive sample experience on that.

There's really no benefit to making them go to latest.

robinsh:
Does that mean you want me to keep the one with

+    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.*.0" />

And remove the one with

-    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.*" />

(just double-checking. The last sdk guy made me change it to be the other way. And I don’t care, I just want it to be the way you would want it to be.”

Robin

robinsh:
No, that was just a note to myself of the name of the sample. Oops.

robinsh:
Can you take pity on me and save me from searching the c# reference and figuring how to make a using statement to be used here? There’s another one, too, I’m about to send. And I think this will take care of my c# routing tutorial issues.

I do want to ask an implementation q about the nuget versioning. So your predecessor told me to always use the last nuger version, including an asterisk if necessary, thus ensuring it would be the most recent version. But you told me to always use the latest “real” version (Which I did). Why would you rather have it set to 1.26.0 instead 1.26.*? (Jimaco asked). You’re not going to go update all of them through the samples when a newer version comes out, are you?

Thanks (again) for your help.

Robijn

robinsh:
Last question – did you throw out my token cancellation or my while (true)? Because I want the user to be able to hit return to stop the loop, rather than having it run infinitely.

iot-hub/Tutorials/Routing/SimulatedDevice/Program.cs
drwill-ms: `using var cts = new CancellationTokenSource();`

If that doesn't work, (there is a version requirement), you can give it a body:

```csharp
using (var cts = new CancellationTokenSource())
{
    /// ...
}
```

robinsh:
hey @drwill-ms I think I've got this fixed now. I'm going to sign off on it. Let me know if I missed anything.

iot-hub/Tutorials/Routing/SimulatedDevice/Program.cs
```diff
@@ -33,16 +29,16 @@ class Program
     {
         private static DeviceClient s_deviceClient;
         private readonly static string s_myDeviceId = "Contoso-Test-Device";
-        private readonly static string s_iotHubUri = "ContosoTestHubdlxlud5h.azure-devices.net";
+        private readonly static string s_iotHubUri = "<iot-hub-name-goes-here>.azure-devices.net";
         // This is the primary key for the device. This is in the portal. 
         // Find your IoT hub in the portal > IoT devices > select your device > copy the key. 
-        private readonly static string s_deviceKey = "RClD0LGxZCYavagk8tS2M7L1MI5bcKcyR+tJHzj+gDk=";
```
drwill-ms, line 39: If you didn't already, be sure to cycle this key

drwill-ms (approved):

robinsh:
#sign-off

PR #134, opened 2020-09-23
Add a Thermostat and TemperatureController samples to illustrate service-side operations on a plug and play device

barustum:
## Purpose

Add a Thermostat and TemperatureController samples to illustrate how to perform root level and component level service-side operations on a plug and play compatible device.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: Adding new sample
```

### DIFF
--- a/iot-hub/Samples/service/DigitalTwinClientSamples/TemperatureController/TemperatureController.csproj
+++ b/iot-hub/Samples/service/DigitalTwinClientSamples/TemperatureController/TemperatureController.csproj
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.8.0" />
-    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.1-pnp-preview" />
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.23.0-pnp-preview" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="3.1.8" />
   </ItemGroup>
 
--- a/iot-hub/Samples/service/DigitalTwinClientSamples/TemperatureController/TemperatureController.csproj
+++ b/iot-hub/Samples/service/DigitalTwinClientSamples/TemperatureController/TemperatureController.csproj
@@ -1,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.8.0" />
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.1-pnp-preview" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="3.1.8" />
+  </ItemGroup>
+
+</Project>
--- a/iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Parameters.cs
+++ b/iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Parameters.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using CommandLine;
+
+namespace Microsoft.Azure.Devices.Samples
+{
+    /// <summary>
+    /// Parameters for the application supplied via command line arguments.
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            'c',
+            "HubConnectionString",
+            Required = true,
+            HelpText = "The IoT Hub connection string. This is available under the \"Shared access policies\" in the Azure portal.")]
+        public string HubConnectionString { get; set; }
+
+        [Option(
+            'd',
+            "DigitalTwinId",
+            Required = true,
+            HelpText = "The Id of the plug and play compatible device to be used in the sample.")]
+        public string DigitalTwinId { get; set; }
+    }
+}
--- a/iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Parameters.cs
+++ b/iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Parameters.cs
@@ -2,26 +2,45 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using CommandLine;
+using System;
 
 namespace Microsoft.Azure.Devices.Samples
 {
     /// <summary>
     /// Parameters for the application supplied via command line arguments.
+    /// If the parameter is not supplied via command line args, it will look for it in environment variables.
     /// </summary>
     internal class Parameters
     {
         [Option(
             'c',
             "HubConnectionString",
-            Required = true,
-            HelpText = "The IoT Hub connection string. This is available under the \"Shared access policies\" in the Azure portal.")]
-        public string HubConnectionString { get; set; }
+            HelpText = "The IoT Hub connection string. This is available under the \"Shared access policies\" in the Azure portal." +
+            "\nDefaults to environment variable \"IOTHUB_CONNECTION_STRING\".")]
+        public string HubConnectionString { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_CONNECTION_STRING");
 
         [Option(
             'd',
-            "DigitalTwinId",
-            Required = true,
-            HelpText = "The Id of the plug and play compatible device to be used in the sample.")]
-        public string DigitalTwinId { get; set; }
+            "DeviceId",
+            HelpText = "The Id of the plug and play compatible device to be used in the sample." +
+            "\nDefaults to environment variable \"IOTHUB_DEVICE_ID\"")]
+        public string DeviceId { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_ID");
+
+        public bool Validate()
+        {
+            if (string.IsNullOrWhiteSpace(HubConnectionString))
+            {
+                throw new ArgumentNullException(nameof(HubConnectionString), "An IoT Hub connection string needs to be specified, please set the environment variable \"IOTHUB_CONNECTION_STRING\"" +
+                    "or pass in \"-c | --HubConnectionString\" through command line.");
+            }
+
+            if (string.IsNullOrWhiteSpace(DeviceId))
+            {
+                throw new ArgumentNullException(nameof(DeviceId), "The Id of the plug and play device needs to be specified, please set the environment variable \"IOTHUB_DEVICE_ID\"" +
+                    "or pass in \"-d | --DeviceId\" through command line.");
+            }
+
+            return true;
+        }
     }
 }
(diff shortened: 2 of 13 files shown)


iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        /// <summary>
+        /// A sample to illustrate how to perform root level service-side operations on a plug and play compatible device.
+        /// </summary>
+        /// <param name="args">
+        /// Run with `--help` to see a list of required and optional parameters.
+        /// </param>
+        /// <remarks>
+        /// This sample performs operations on a plug and play compatible device on the root level. For a sample on how to perform
+        /// operations on a component level on a plug an dplay compatible device, please check out <see href="https://github.com/Azure-Samples/azure-iot-samples-csharp/tree/master/iot-hub/Samples/device">PnpDeviceSample</see>.
```
barustum, line 23: TODO: the href link here needs to be updated once the device sample place is finalized.

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Parameters.cs
```diff
@@ -0,0 +1,27 @@
...
+            Required = true,
+            HelpText = "The IoT Hub connection string. This is available under the \"Shared access policies\" in the Azure portal.")]
+        public string HubConnectionString { get; set; }
+
+        [Option(
+            'd',
+            "DigitalTwinId",
+            Required = true,
+            HelpText = "The Id of the plug and play compatible device to be used in the sample.")]
+        public string DigitalTwinId { get; set; }
```
abhipsaMisra, line 25: nit: we'll need these to default to env variables, something like this: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/samples/DigitalTwinDeviceSamples/Thermostat/Parameter.cs#L20

Env var names:
IOTHUB_CONNECTION_STRING
IOTHUB_DEVICE_ID

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Program.cs
abhipsaMisra, line 23: "... operations on a component level on a plug an dplay compatible device ..."
should this be component-level?

typo on play

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,64 @@
...
+                {
+                    Environment.Exit(1);
+                });
+
+            s_logger = InitializeConsoleDebugLogger();
+
+            s_logger.LogDebug("Set up the digital twin client.");
+            using DigitalTwinClient digitalTwinClient = DigitalTwinClient.CreateFromConnectionString(parameters.HubConnectionString);
+
+            s_logger.LogDebug("Set up and start the Thermostat sample.");
```
abhipsaMisra, line 44: nit: maybe call it out that we are starting the thermostat service sample? so as to not confuse it with the device sample

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/ThermostatSample.cs
```diff
@@ -0,0 +1,136 @@
...
+            // Get and print the digital twin
+            await GetAndPrintDigitalTwin<ThermostatTwin>();
+
+            // Update the targetTemperature property of the digital twin
+            await UpdateTargetTemperatureProperty();
+
+            // Add, replace then remove currentTemperature property on the digital twin
+            await UpdateCurrentTemperatureProperty();
+
+            // Invoke the getMaxMinReport command on the root level of the digital twin
```
abhipsaMisra, line 38: nit: instead of saying - "invoke the command {abc} on the root level of the digital twin", how do you feel about saying "invoke the root-level command {abc} on the digital twin".

I think the 2nd one better highlights that we are invoking a root-level command.

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/ThermostatSample.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        private async Task UpdateTargetTemperatureProperty()
+        {
+            const string targetTemperaturePropertyName = "targetTemperature";
+            var updateOperation = new UpdateOperationsUtility();
+
+            // Choose a random value to assign to the targetTemperature property
+            int desiredTargetTemperature = Random.Next(0, 100);
+
+            // First let's take a look at when the property was updated and what was it set to.
```
abhipsaMisra, line 61: nice!

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/ThermostatSample.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            _logger.LogDebug($"Update {_digitalTwinId} digital twin response: {updateDigitalTwinResponse.Response.StatusCode}.");
+        }
+
+        private async Task UpdateCurrentTemperatureProperty()
+        {
+            // Choose a random value to assign to the currentTemperature property
+            int currentTemperature = Random.Next(0, 100);
+
+            const string currentTemperaturePropertyName = "currentTemperature";
```
abhipsaMisra, line 92: We should call out explicitly that this property doesn't exist on model that this device has "registered" with

abhipsaMisra: Just to refresh my memory, why did we decide against doing these operations on the root level "targetTemperature" directly?

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/ThermostatSample.cs
```diff
@@ -0,0 +1,136 @@
...
+            await GetAndPrintDigitalTwin<ThermostatTwin>();
+
+            // Third, remote the currentTemperature property
+            updateOperation.AppendRemoveOp($"/{currentTemperaturePropertyName}");
+            _logger.LogDebug($"Remove the {currentTemperaturePropertyName} property on the {_digitalTwinId} digital twin.");
+            HttpOperationHeaderResponse<DigitalTwinUpdateHeaders> removePropertyInDigitalTwinResponse = await _digitalTwinClient.UpdateDigitalTwinAsync(_digitalTwinId, updateOperation.Serialize());
+            _logger.LogDebug($"Update {_digitalTwinId} digital twin response: {removePropertyInDigitalTwinResponse.Response.StatusCode}.");
+            // Print the Thermostat digital twin
+            await GetAndPrintDigitalTwin<ThermostatTwin>();
+        }
```
abhipsaMisra, line 119: super nitty: what do you think about adding something that shows how to do multiple patch operations through a single Update call?
meaning, the updateOperations will have both a replace and an add (could be for different properties), but just to highlight that the `Update` API takes in a list of update operations => you can pass in more than one.

This might be a better idea for the temperature controller one -> we have multiple components with multiple properties available to us there.
Thoughts?

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/ThermostatTwin.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Azure.Devices.Serialization;
+using Newtonsoft.Json;
+using System;
+
+namespace Microsoft.Azure.Devices.Samples
+{
+    public class ThermostatTwin : BasicDigitalTwin
```
abhipsaMisra, line 10: I have updated this class in https://github.com/Azure/azure-iot-sdk-csharp/pull/1572.
You can update the same here as well.

Notably, I am not using inheritance from the BasicDigitalTwin etc types for these custom twins.

abhipsaMisra (approved):
Some minor comments, looks good otherwise!

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Parameters.cs
barustum: Done

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,64 @@
...
+        private static ILogger s_logger;
+
+        /// <summary>
+        /// A sample to illustrate how to perform root level service-side operations on a plug and play compatible device.
+        /// </summary>
+        /// <param name="args">
+        /// Run with `--help` to see a list of required and optional parameters.
+        /// </param>
+        /// <remarks>
+        /// This sample performs operations on a plug and play compatible device on the root level. For a sample on how to perform
```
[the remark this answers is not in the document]
barustum: Removed the remarks section as the readme takes care of illustrating this point.

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/ThermostatSample.cs
barustum: I'll add a comment to explicitly call out that this property does not exist on the model registered with this device.

Regarding your question, I don't think we discussed why we can't do these operations on targetTemperature. We can. But I feel like I don't want to remove targetTemperature property because it's in the model. So if the user wants to run this sample again, against the same device id, that targetTemperature won't exist anymore because it's been deleted due to the previous run of the sample.

barustum: Added a comment to the RunSampleAsync to explicitly make the distinction about currentTemperature

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/ThermostatSample.cs
barustum: Good idea. I will leave this comment open to get to it after all higher priority items are resolved.

iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/ThermostatTwin.cs
barustum: Updated

iot-hub/Samples/service/DigitalTwinClientSamples/TemperatureController/TemperatureController.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.8.0" />
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.1-pnp-preview" />
```
abhipsaMisra, line 10: just fyi - private nuget versions here: https://github.com/Azure/azure-iot-sdk-csharp/pull/1576

abhipsaMisra (approved):
Looks good!

iot-hub/Samples/service/DigitalTwinClientSamples/TemperatureController/TemperatureController.csproj
barustum: updated to 1.23.0-pnp-preview

PR #138, opened 2020-09-24
Fix error with Message body being re-read, and loop termination

drwill-ms:


iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -194,15 +194,15 @@ private async Task SendMessagesAsync(CancellationToken cancellationToken)
...
 
-                    Message message = PrepareMessage(messageCount);
+                    (Message message, string payload) = PrepareMessage(messageCount);
                     while (true)
                     {
                         try
                         {
                             await s_deviceClient.SendEventAsync(message);
-                            _logger.LogInformation($"Sent message {messageCount} with data [{message.GetBytes()}]");
+                            _logger.LogInformation($"Sent message {messageCount} of {payload}");
```
drwill-ms, line 203: Can't read the message bytes again after having sent it (when it was read), so have `PrepareMessage` return the payload as well.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -194,15 +194,15 @@ private async Task SendMessagesAsync(CancellationToken cancellationToken)
...
                             message.Dispose();
-                            continue;
+                            break;
```
drwill-ms, line 205: wrong loop termination

abhipsaMisra (approved):

PR #148, opened 2020-10-03
Add lunchSettings with EnvVars predefined

rido-min:
## Purpose

The PnP Samples require some environment variables to work.
This PR includes lunchSettings for VS and VSCode with the variable names already pre-configured.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe:
```

## How to Test
*  Get the code

image

image

### DIFF
--- a/iot-hub/Samples/device/PnpDeviceSamples/.vscode/launch.json
+++ b/iot-hub/Samples/device/PnpDeviceSamples/.vscode/launch.json
@@ -1,0 +1,74 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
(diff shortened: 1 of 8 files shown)


drwill-ms (changes requested):
I like that this will make it easier for someone to run the samples with VS Code.

The danger of someone editing these files with their secrets and accidentally pushing a branch or requesting a PR with the changes is too high.

So instead, let's ensure our .gitignore file excludes these file names, rename these files to something that can be added, and add instructions in a readme to copy the file locally to the right name.

rido-min:
I've renamed the files to .template.json 
Also note that this is not only for VSCode, it also applies to VS.

PR #151, opened 2020-10-05
Consolidate module sample into one, and add reconnection logic

abhipsaMisra:
Link to readme: https://github.com/Azure-Samples/azure-iot-samples-csharp/tree/abmisr/moduleSample/iot-hub/Samples/module/ModuleSample

The reconnection logic and sample structure is the same as the DeviceReconnectionSample. I've added some additional setup info and code snippet into the readme (to help people who simply want to check the API usage, and not a complete module simulation).


iot-hub/Samples/module/ModuleSample/ExceptionExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        internal static IEnumerable<Exception> Unwind(this Exception exception, bool unwindAggregate = false)
+        {
+            while (exception != null)
+            {
+                yield return exception;
+
+                if (!unwindAggregate)
+                {
+                    exception = exception.InnerException;
```
drwill-ms, line 19: does this change which exception the original object object points to?

iot-hub/Samples/module/ModuleSample/readme.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+### Send telemetry:
+
+```csharp
+var temperature = 25;
+var humidity = 70;
+string messagePayload = $"{{\"temperature\":{temperature},\"humidity\":{humidity}}}";
+
+var eventMessage = new Message(Encoding.UTF8.GetBytes(messagePayload))
```
drwill-ms, line 36: using

iot-hub/Samples/module/ModuleSample/readme.md
```diff
@@ -0,0 +1,135 @@
...
+{
+    MessageId = Guid.NewGuid().ToString(),
+    ContentEncoding = Encoding.UTF8.ToString(),
+    ContentType = "application/json",
+};
+
+await moduleClient.SendEventAsync(message);
```
drwill-ms, line 43: worth talking about which classes of exceptions one might expect, and what app behavior is required based on that. Meaning, you might retry, reinit your client, or quit.

drwill-ms:
What samples were consolidated?

drwill-ms:
I lament we copy code from another sample, meaning we'll need to update it in multiple places. What do you think about making a sample helper library for common stuff?

iot-hub/Samples/module/ModuleSample/ExceptionExtensions.cs
abhipsaMisra: no, it simply returns the inner exceptions for the exception, so that we can use them for verification/ mapping etc
eg case - where the outer exception is IotHubCommunicationException, but the inner exception is a TimeoutException -> the client should retry during operation timeouts.

iot-hub/Samples/module/ModuleSample/readme.md
abhipsaMisra: Good point, how about I add a link to our sample where we handle the most common cases?

abhipsaMisra:
> What samples were consolidated?

The module messaging sample, and the module twin sample.

iot-hub/Samples/module/ModuleSample/Logging/ColorConsoleLogger.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// For additional details, see https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger?view=dotnet-plat-ext-3.1.
+    /// </summary>
+    public class ColorConsoleLogger : ILogger
+    {
+        private readonly ColorConsoleLoggerConfiguration _config;
+
+        /// <summary>
+        /// Initializes an instance of <see cref="ColorConsoleLogger"/>.
+        /// </summary>
+        /// <param name="name">The category name for each <see cref="ILogger"/> instance. This is usually the class name where the logger is initialized.</param>
```
barustum, line 22: no name param

abhipsaMisra:
> I lament we copy code from another sample, meaning we'll need to update it in multiple places. What do you think about making a sample helper library for common stuff?

I totally agree, although, we might be better off consolidating it as a part of the "samples-refresh", what do you think?

- We can take in the messaging code, the twin update code, etc as sub-modules; and refer to it in both the samples.
- We can create a pluggable client, which performs these operations, and pass in a device client/ module client as required.

I think we might come up with more ideas as we clean up the samples, and that might help us in getting to a better (more efficient/ cleaner) approach.

iot-hub/Samples/module/ModuleSample/Logging/ColorConsoleLogger.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            if (!IsEnabled(logLevel))
+            {
+                return;
+            }
+
+            var color = _config.LogLevelToColorMapping[logLevel];
+            if (_config.EventIds.Contains(ColorConsoleLoggerConfiguration.DefaultEventId) || _config.EventIds.Contains(eventId.Id))
+            {
+                var initialColor = Console.ForegroundColor;
```
barustum, line 67: nit: ConsoleColor instead of var.

iot-hub/Samples/module/ModuleSample/Logging/ColorConsoleLogger.cs
abhipsaMisra: whoops, good catch. This is argument enough to separate out the logger into its own project, I'll do that now.

iot-hub/Samples/module/ModuleSample/readme.md
drwill-ms: Elaborate?

drwill-ms (approved):

iot-hub/Samples/module/ModuleSample/readme.md
abhipsaMisra: We touch on those in the section "Module reconnection" below. I feel that if we add those instructions here again, it might be repetitive. 
I don't necessarily want to list exceptions based on operation - i.e. you might see _these_ exceptions while sending telemetry - since the list might not be exhaustive (we'd don't have an operation: exception mapping yet).

Thoughts?

iot-hub/Samples/module/ModuleSample/ModuleSample.cs
```diff
@@ -0,0 +1,265 @@
...
+        private static ConnectionStatus s_connectionStatus;
+        private static bool s_wasEverConnected;
+
+        public ModuleSample(List<string> moduleConnectionStrings, TransportType transportType, ILogger logger)
+        {
+            _logger = logger;
+
+            // This class takes a list of potentially valid connection strings (most likely the currently known good primary and secondary keys)
+            // and will attempt to connect with the first. If it receives feedback that a connection string is invalid, it will discard it, and
+            // if any more are remaining, will try the next one.
```
barustum, line 40: I believe 'any more' should be 'anymore'

iot-hub/Samples/module/ModuleSample/ModuleSample.cs
```diff
@@ -0,0 +1,265 @@
...
+                                " you should dispose (DisposeAsync()) and then open (OpenAsync()) the client.");
+
+                            InitializeClient();
+                            break;
+
+                        default:
+                            _logger.LogError("### This combination of ConnectionStatus and ConnectionStatusChangeReason is not expected," +
+                                " contact the client library team with logs.");
+                            break;
+
```
barustum, line 182: extra line

barustum (approved):

iot-hub/Samples/module/ModuleSample/ModuleSample.cs
abhipsaMisra: any more = some more, quantitatively = would you like any more apples; 
anymore = relating to time = I don't like apples anymore.

so in this context, it should be "any more" => any more connection strings are remaining...

PR #158, opened 2020-11-11
Update reconnection sample to use latest sdk and update reconnection logic

abhipsaMisra:


iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -49,8 +50,6 @@ public DeviceReconnectionSample(List<string> deviceConnectionStrings, TransportT
 
             _transportType = transportType;
             _logger.LogInformation($"Using {_transportType} transport.");
-
-            InitializeClient();
```
abhipsaMisra, line 53: calling this within `RunSampleAsync`, it doesn't need to be in the constructor necessarily.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -96,24 +96,17 @@ private void InitializeClient()
...
                     _logger.LogDebug($"Initialized the client instance.");
                 }
 
                 try
                 {
                     // Force connection now.
                     // OpenAsync() is an idempotent call, it has the same effect if called once or multiple times on the same client.
-                    s_deviceClient.OpenAsync().GetAwaiter().GetResult();
-                    _logger.LogDebug($"Initialized the client instance.");
+                    await s_deviceClient.OpenAsync();
```
abhipsaMisra, line 108: updated this method to be async

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -122,11 +115,18 @@ private void InitializeClient()
             }
         }
 
-        private void ConnectionStatusChangeHandler(ConnectionStatus status, ConnectionStatusChangeReason reason)
+        // It is not good practice to have async void methods, however, DeviceClient.SetConnectionStatusChangesHandler() event handler signature has a void return type.
+        // As a result, any operation within this block will be executed unmonitored on another thread.
+        // To prevent multi-threaded synchronization issues, the async method InitializeClientAsync being called in here first grabs a lock
+        // before attempting to initailize or dispose the device client instance.
+        private async void ConnectionStatusChangeHandler(ConnectionStatus status, ConnectionStatusChangeReason reason)
```
abhipsaMisra, line 122: this is definitely a bad pattern, but that is because of the way `ConnectionStatusChangeHandler` is defined. We have locks within the async method `InitializeClientAsync`, to ensure that there are no multi-threaded issues.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.csproj
```diff
@@ -9,7 +9,7 @@
 
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.8.0" />
-    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.33.0" />
+    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.33.1" />
```
abhipsaMisra, line 12: this version of the sdk gives us the correct status callback on initial disconnect (so we don't need the `wasEverConnceted` etc. flag anymore).

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
```diff
@@ -160,7 +160,7 @@ The device client exhibits the following connection status changes with reason:
   <tr>
     <td> Retry_Expired </td>
     <td> This signifies that the client was disconnected due to a transient exception, but the retry policy expired before a connection could be re-established. </td>
-    <td> If you want to perform more operations on the device client, you should dispose and then re-initialize the client. </td>
+    <td> If you want to perform more operations on the device client, you should dispose and then re-initialize the client. </br> Note that the SDK's default retry policy is set to never expire. </td>
```
timtay-microsoft, line 163: Does the 4 minute default operation timeout not apply here?

abhipsaMisra: The retry policy won't expire, your operation will time out -> the client will still be in connected state. With the retry policy expiring, the client would go into disconnected state.

abhipsaMisra: Actually let me test it out, it'll be really bad if I put it into the doc and it turns out to be incorrect.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -122,11 +115,18 @@ private void InitializeClient()
...
         {
-            _logger.LogDebug($"Connection status changed: status={status}, reason={reason}");
+            lock (_connectionStatusLock)
+            {
+                _logger.LogDebug($"Connection status changed: status={status}, reason={reason}");
+                s_connectionStatus = status;
```
drwill-ms, line 127: If you are locking to avoid it being overwritten in another thread, you still have the problem of the value changing from here to the switch statement.

My feeling is there is no value in locking here. Just use volatile, or in the switch statement use `status` not the member variable.

drwill-ms (approved):

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
abhipsaMisra: Yeah, we get a task cancelled exception - the retry policy is still active:
```
2020-11-11T15:07:50.8091249>> Debug - Connection status changed: status=Disconnected_Retrying, reason=Communication_Error
2020-11-11T15:07:50.8170451>> Debug - ### The DeviceClient is retrying based on the retry policy. Do NOT close or open the DeviceClient instance
2020-11-11T15:07:51.2935904>> Error - A network related exception was caught, but will try to recover and retry explicitly: System.TimeoutException: Timed out to acquire handler lock.
   at Microsoft.Azure.Devices.Client.Transport.RetryDelegatingHandler.EnsureOpenedAsync(TimeoutHelper timeoutHelper)
   at Microsoft.Azure.Devices.Client.Transport.RetryDelegatingHandler.<>c__DisplayClass19_0.<<ReceiveAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.Azure.Devices.Client.Transport.RetryDelegatingHandler.ReceiveAsync(TimeoutHelper timeoutHelper)
   at Microsoft.Azure.Devices.Client.InternalClient.ReceiveAsync(TimeSpan timeout)
   at Microsoft.Azure.Devices.Client.Samples.DeviceReconnectionSample.ReceiveMessageAndCompleteAsync() in C:\MyData\SourceCode\csharp\temp\azure-iot-samples-csharp\iot-hub\Samples\device\DeviceReconnectionSample\DeviceReconnectionSample.cs:line 293
   at Microsoft.Azure.Devices.Client.Samples.DeviceReconnectionSample.ReceiveMessagesAsync(CancellationToken cancellationToken) in C:\MyData\SourceCode\csharp\temp\azure-iot-samples-csharp\iot-hub\Samples\device\DeviceReconnectionSample\DeviceReconnectionSample.cs:line 256
2020-11-11T15:08:01.7451503>> Error - A network related exception was caught, but will try to recover and retry: System.TimeoutException: The operation timed out.
 ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at Microsoft.Azure.Devices.Client.Transport.RetryDelegatingHandler.SendEventAsync(Message message, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.InternalClient.SendEventAsync(Message message)
   --- End of inner exception stack trace ---
   at Microsoft.Azure.Devices.Client.InternalClient.SendEventAsync(Message message)
   at Microsoft.Azure.Devices.Client.Samples.DeviceReconnectionSample.SendMessagesAsync(CancellationToken cancellationToken) in C:\MyData\SourceCode\csharp\temp\azure-iot-samples-csharp\iot-hub\Samples\device\DeviceReconnectionSample\DeviceReconnectionSample.cs:line 213
2020-11-11T15:08:36.8624005>> Error - A network related exception was caught, but will try to recover and retry: System.TimeoutException: The operation timed out.
 ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at Microsoft.Azure.Devices.Client.Transport.RetryDelegatingHandler.SendEventAsync(Message message, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.InternalClient.SendEventAsync(Message message)
   --- End of inner exception stack trace ---
   at Microsoft.Azure.Devices.Client.InternalClient.SendEventAsync(Message message)
   at Microsoft.Azure.Devices.Client.Samples.DeviceReconnectionSample.SendMessagesAsync(CancellationToken cancellationToken) in C:\MyData\SourceCode\csharp\temp\azure-iot-samples-csharp\iot-hub\Samples\device\DeviceReconnectionSample\DeviceReconnectionSample.cs:line 213
2020-11-11T15:08:47.8606058>> Debug - Connection status changed: status=Connected, reason=Connection_Ok
2020-11-11T15:08:47.8612025>> Debug - ### The DeviceClient is CONNECTED; all operations will be carried out as normal.
2020-11-11T15:08:48.3162341>> Information - Sent message 3 of {"temperature":29,"humidity":64}
```

This was with an operation timeout of 30secs, disconnection of about 1 min.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -154,7 +154,7 @@ private void ConnectionStatusChangeHandler(ConnectionStatus status, ConnectionSt
                             {
                                 // Not great to print out a connection string, but this is done for sample/demo purposes.
                                 _logger.LogWarning($"The current connection string {badCs} is invalid. Trying another.");
-                                InitializeClient();
+                                await InitializeAndOpenClientAsync();
```
vinagesh, line 157: Why do we need InitializeAndOpenClientAsync at all here? It seems a little circular to set the ConnectionStatusChangeHandler within InitializeAndOpenClientAsync and call InitializeAndOpenClientAsync inside ConnectionStatusChangeHandler.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -154,7 +154,7 @@ private void ConnectionStatusChangeHandler(ConnectionStatus status, ConnectionSt
                             {
                                 // Not great to print out a connection string, but this is done for sample/demo purposes.
```
vinagesh, line 155: I don't think we should print out the connection string even if it is a sample and we have a statement that says it is not a good idea.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: Yeah, I was counting on it being fast enough to not matter, but that will not be true for all cases. I could put the entire switch block inside the lock, but that seemed to be an overkill. I like the idea of using the method arg `status` directly in the switch, I'll do that.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -170,14 +170,14 @@ private void ConnectionStatusChangeHandler(ConnectionStatus status, ConnectionSt
                             _logger.LogWarning("### The DeviceClient has been disconnected because the retry policy expired." +
                                 "\nIf you want to perform more operations on the device client, you should dispose (DisposeAsync()) and then open (OpenAsync()) the client.");
 
-                            InitializeClient();
+                            await InitializeAndOpenClientAsync();
                             break;
 
                         case ConnectionStatusChangeReason.Communication_Error:
                             _logger.LogWarning("### The DeviceClient has been disconnected due to a non-retry-able exception. Inspect the exception for details." +
```
vinagesh, line 177: I know this is already there but change to string interpolation?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: I see your point, but in the absence of a "IsClientConnected" property on the DeviceClient, we'll have to manage the device client's lifecycle from within the status change callback. So the flow becomes:
-> initialize client -> set connection status callback handler -> use connection status callback handler to detect disconnection -> dispose the client -> initialize client (back to the start) -> ...

Do you think there might be a better way to handle this? Our current client indicates the status of operation through exceptions, and the status of the client through event handler. This does make managing the lifecycle confusing, but this is the approach I found that worked. If we can come up with something that'll be easier to implement, I would like that.

vinagesh: I think I get why we are doing this. However, we are not really calling dispose before calling open right? Or am I missing something? Is it a common patter to initialize the client on error within the ConnectionStatusChangeHandler ?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: I hear ya, I'll remove the connection string from the log statement.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
vinagesh: Ok I found the dispose :) Took a while to go through the flow.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: This is just a long string broken into two links, string interpolation wouldn't really help here.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
vinagesh: If you are expecting multiple threads to access this, it could cause lots of other issues. Ex: You are trying to initialize the client inside the switch but there is another thread that sets the s_connectionStatus  to something else, causing the init to skip completely.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
vinagesh: Ideally, it would be good for the DeviceClient itself to have a property to know if we need to reinitialize it. It might be better to just store s_connectionStatus and use that during Init. We don't need to call the Init here. Also if we are expecting multiple threads to hit this, the end result should be like an AND or OR of all the status so that something is not overriden.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: but that is what we want, we don't initialize s_connectionStatus anywhere other than within this callback, so if the callback updated s_connectionStatus to be Connected, we want the init block to be skipped. For this reason, we have a check for the status s_connectionStatus inside the init block as well.

vinagesh: And if threads can re-write this value in any order, we might be losing some info on s_connectionStatus. It should be an OR or AND based on the logic we use to store the result of all callbacks.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
vinagesh: Ahhh I see!

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: > We don't need to call the Init here.

Where would we handle the device client's lifetime then? The only monitoring tool that that device client gives you is this status change callback.

> the end result should be like an AND or OR of all the status so that something is not overridden.

I didn't understand what you meant here, could you give an example?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -18,11 +18,12 @@ public class DeviceReconnectionSample
         private const int TemperatureThreshold = 30;
 
         private static readonly TimeSpan s_sleepDuration = TimeSpan.FromSeconds(5);
-        private static readonly TimeSpan s_operationTimeout = TimeSpan.FromHours(1);
 
         private readonly object _initLock = new object();
+        private readonly object _connectionStatusLock = new object();
         private readonly List<string> _deviceConnectionStrings;
         private readonly TransportType _transportType;
+        private readonly ClientOptions _clientOptions = new ClientOptions { SdkAssignsMessageId = Shared.SdkAssignsMessageId.WhenUnset };
```
bikamani, line 26: SdkAssignsMessageId ? What does this indicate? I'd rename for readability

vinagesh (approved):

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: It means that if you do not set a message Id on a message, the the sdk will assign one for you

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
vinagesh: Talked offline. We do think there is a better solution overall for this scenario but we'll do it when we re-design the device client.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
vinagesh: Talked offline

PR #160, opened 2020-11-20
change tempcontroller sample to v2 model

ericmitt:
## Purpose
Make this sample compatible with IoT Central
* ...

```
[X] No
```

## Pull Request Type
Make this sample compatible with IoT Central

```
[ ] Bugfix
[X ] Feature
[ ] Other... Please describe:
```

## How to Test
Run the sample as usual, it now use the v2:
private const string ModelId = "dtmi:com:example:TemperatureController;2";

### DIFF
--- a/iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Models/TemperatureController.json
+++ b/iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Models/TemperatureController.json
@@ -1,58 +1,83 @@
 {
-  "@context": "dtmi:dtdl:context;2",
-  "@id": "dtmi:com:example:TemperatureController;1",
-  "@type": "Interface",
-  "displayName": "Temperature Controller",
-  "description": "Device with two thermostats and remote reboot.",
-  "contents": [    {
-      "@type": [
-        "Telemetry", "DataSize"
-      ],
-      "name": "workingSet",
-      "displayName": "Working Set",
-      "description": "Current working set of the device memory in KiB.",
-      "schema": "double",
-      "unit" : "kibibyte"
-    },
-    {
-      "@type": "Property",
-      "name": "serialNumber",
-      "displayName": "Serial Number",
-      "description": "Serial number of the device.",
-      "schema": "string"
-    },
-    {
-      "@type": "Command",
-      "name": "reboot",
-      "displayName": "Reboot",
-      "description": "Reboots the device after waiting the number of seconds specified.",
-      "request": {
-        "name": "delay",
-        "displayName": "Delay",
-        "description": "Number of seconds to wait before rebooting the device.",
-        "schema": "integer"
+    "@context": [
+      "dtmi:iotcentral:context;2",
+      "dtmi:dtdl:context;2"
+    ],
+    "@id": "dtmi:com:example:TemperatureController;2",
+    "@type": "Interface",
+    "contents": [
+      {
+        "@type": [
+          "Telemetry",
+          "DataSize"
+        ],
+        "description": {
+          "en": "Current working set of the device memory in KiB."
+        },
+        "displayName": {
+          "en": "Working Set"
+        },
+        "name": "workingSet",
+        "schema": "double",
+        "unit": "kibibit"
+      },
+      {
+        "@type": "Property",
+        "displayName": {
+          "en": "Serial Number"
+        },
+        "name": "serialNumber",
+        "schema": "string",
+        "writable": false
+      },
+      {
+        "@type": "Command",
+        "commandType": "synchronous",
+        "description": {
+          "en": "Reboots the device after waiting the number of seconds specified."
+        },
+        "displayName": {
+          "en": "Reboot"
+        },
+        "name": "reboot",
+        "request": {
+          "@type": "CommandPayload",
+          "description": {
+            "en": "Number of seconds to wait before rebooting the device."
+          },
+          "displayName": {
+            "en": "Delay"
+          },
+          "name": "delay",
+          "schema": "integer"
+        }
+      },
+      {
+        "@type": "Component",
+        "displayName": {
+          "en": "thermostat1"
+        },
+        "name": "thermostat1",
+        "schema": "dtmi:com:example:Thermostat;1"
+      },
+      {
+        "@type": "Component",
+        "displayName": {
+          "en": "thermostat2"
+        },
+        "name": "thermostat2",
+        "schema": "dtmi:com:example:Thermostat;2"
+      },
+      {
+        "@type": "Component",
+        "displayName": {
+          "en": "DeviceInfo"
+        },
+        "name": "deviceInformation",
+        "schema": "dtmi:azure:DeviceManagement:DeviceInformation;1"
       }
-    },
-    {
-      "@type" : "Component",
-      "schema": "dtmi:com:example:Thermostat;1",
-      "name": "thermostat1",
-      "displayName": "Thermostat One",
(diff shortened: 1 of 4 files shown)


rido-min (changes requested):
Now that we a have a device model repo I dont think we need to copy the json files anymore here, instead update the link to: https://devicemodels.azure.com/dtmi/com/example/temperaturecontroller-2.json or https://devicemodels.azure.com/dtmi/com/example/temperaturecontroller-2.expanded.json

Wondering if we need to make any update to the service side samples

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Models/TemperatureController.json
```diff
@@ -1,58 +1,83 @@
...
+        "name": "thermostat1",
+        "schema": "dtmi:com:example:Thermostat;1"
+      },
+      {
+        "@type": "Component",
+        "displayName": {
+          "en": "thermostat2"
+        },
+        "name": "thermostat2",
+        "schema": "dtmi:com:example:Thermostat;2"
```
barustum, line 69: Before, the schema for thermostat2 was using dtmi:com:example:Thermostat;1. Don't think this will break the sample, but wondering if you ran the sample to double check.

ericmitt: I tested the service sample, no changes needed

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -15,8 +15,8 @@ namespace Microsoft.Azure.Devices.Client.Samples
 {
     public class Program
     {
-        // DTDL interface used: https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/TemperatureController.json
-        private const string ModelId = "dtmi:com:example:TemperatureController;1";
+        // DTDL interface used: https://github.com/Azure/iot-plugandplay-models
```
barustum, line 18: nit: maybe point to the actual file https://github.com/Azure/iot-plugandplay-models/blob/main/dtmi/com/example/temperaturecontroller-2.json

ericmitt: what I did in my second commit, do you see it?

rido-min (approved):

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Program.cs
ericmitt: done

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Models/TemperatureController.json
```diff
@@ -1,58 +1,83 @@
...
-      "name": "reboot",
-      "displayName": "Reboot",
-      "description": "Reboots the device after waiting the number of seconds specified.",
-      "request": {
-        "name": "delay",
-        "displayName": "Delay",
-        "description": "Number of seconds to wait before rebooting the device.",
-        "schema": "integer"
+    "@context": [
+      "dtmi:iotcentral:context;2",
```
drwill-ms, line 3: why `iotcentral`?

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Models/TemperatureController.json
drwill-ms, line 69: I think this is going to make for a confusing sample. Customers might wonder:
1. Does my device need to support all versions of an interface?
2. Should my device, under normal circumstances, support 2 versions of an interface?

Why not invent another, distinct interface to support? Just trying to reduce the cost?

drwill-ms (approved):

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Models/TemperatureController.json
bikamani: Fair questions I was wondering about v2 support as well and what it means for customer.

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Models/TemperatureController.json
```diff
@@ -1,58 +1,83 @@
...
+        "name": "thermostat2",
+        "schema": "dtmi:com:example:Thermostat;2"
+      },
+      {
+        "@type": "Component",
+        "displayName": {
+          "en": "DeviceInfo"
+        },
+        "name": "deviceInformation",
+        "schema": "dtmi:azure:DeviceManagement:DeviceInformation;1"
```
bikamani, line 77: Does this remain per v1? Or will it be changing later as per v2?

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Models/TemperatureController.json
bikamani: Do we need to keep v1 models(.json) around?

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -15,8 +15,10 @@ namespace Microsoft.Azure.Devices.Client.Samples
...
+        // in v2 of TempController Model, we use 2 different model of Thermostat,
```
bikamani, line 19: This was confusing to read for me until I looked at the model again carefully. May be good to add information for components something like - TempController model has two components, one following v1 and the second v2 demonstrating that IoT central can handle both model versioning

bikamani (approved):
Added few comments, otherwise looks good to me

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Models/TemperatureController.json
drwill-ms: For this sample to work with Central, it must use two **different** models (differentiated by model Id). That is the only reason for versioning this model (right now).

This is what I talked about the way we're doing this being confusing. It looks like a mistake, rather than an intentional change to make the sample work with Central, given the bug they have. If the two components were of completely different (not versioned) interfaces, then it would appear natural.

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Models/TemperatureController.json
ericmitt: DeviceInfo don't change, stay v1

iot-hub/Samples/device/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
 {
     public class Program
     {
-        // DTDL interface used: https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/Thermostat.json
+        // DTDL interface used: https://github.com/Azure/iot-plugandplay-models/blob/main/dtmi/com/example/temperaturecontroller-2.json
```
barustum, line 17: @ericmitt Did you need to make this comment change? This sample still uses Thermostat model, not temperatureController model

PR #162, opened 2020-12-01
Update DPS device symmetric key sample, and fix warnings in other sln samples

drwill-ms:


provisioning/Samples/device/SymmetricKeySample/Parameters.cs
```diff
@@ -0,0 +1,59 @@
...
+            Required = true,
+            HelpText = "The Id Scope of the DPS instance")]
+        public string IdScope { get; set; }
+
+        [Option(
+            'r',
+            "RegistrationId",
+            Required = true,
+            HelpText = "The Id of the device to register.")]
+        public string RegistrationId { get; set; }
```
timtay-microsoft, line 23: This is only a parameter in individual enrollment scenarios, FYI

drwill-ms: Really it ends up being the device Id. Is it more complicated than I realize?

timtay-microsoft:
In the Java repo, I deliberately split the individual and group enrollment samples for symmetric keys. There is just enough different between the scenarios that I felt it made more sense to split. The two sticking points for me were

- The symmetric key is derived for group enrollments and not individual enrollments

- Group enrollments don't have a "registrationId". They just have a desired device Id. Asking for a registration Id doesn't make sense for group enrollments.

For reference, here are the two java samples:
[individual](https://github.com/Azure/azure-iot-sdk-java/blob/master/provisioning/provisioning-samples/provisioning-symmetrickey-individual-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ProvisioningSymmetricKeyIndividualEnrollmentSample.java)
[group](https://github.com/Azure/azure-iot-sdk-java/blob/master/provisioning/provisioning-samples/provisioning-symmetrickey-group-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/ProvisioningSymmetricKeyEnrollmentGroupSample.java)

provisioning/Samples/device/SymmetricKeySample/ProvisioningDeviceClientSample.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            Console.WriteLine($"Initialized for registration Id {security.GetRegistrationID()}");
+
+            Console.WriteLine("Registering with the device provisioning service... ");
+            DeviceRegistrationResult result = await provClient.RegisterAsync().ConfigureAwait(false);
+
+            Console.WriteLine($"Registration status: {result.Status}.");
+            if (result.Status != ProvisioningRegistrationStatusType.Assigned)
+            {
+                return;
```
barustum, line 56: We return here without telling a user why we stopped the sample. Do we need to print a console message to the user on what the actual status is before returning?

abhipsaMisra, line 56: nit: we could print a message here before returning

provisioning/Samples/device/SymmetricKeySample/SymmetricKeySample.csproj
```diff
@@ -2,15 +2,16 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <LangVersion>8.0</LangVersion>
+    <RootNamespace>Microsoft.Azure.Devices.Provisioning.Client.Samples</RootNamespace>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="..\Common\ProvisioningDeviceClientSample.cs" Link="ProvisioningDeviceClientSample.cs" />
```
abhipsaMisra, line 8: The (provision with dps + test with IoT hub) lived in this common code, with these individual samples showing how to create the different types of security providers. Do we want to deviate from that approach?

barustum (approved):

abhipsaMisra (approved):

provisioning/Samples/device/SymmetricKeySample/Parameters.cs
drwill-ms: Okay, I see that for individual enrollments (in the Azure portal), a registration Id can be different than the device Id.

Working on it...

provisioning/Samples/device/SymmetricKeySample/SymmetricKeySample.csproj
drwill-ms: Yes, I do. Having all the extra code about non-symmetric key for a symmetric key sample is distracting.

timtay-microsoft (approved):

PR #164, opened 2020-12-02
Update device DPS x509 sample

drwill-ms:


barustum (approved):

provisioning/Samples/device/X509Sample/ProvisioningDeviceClientSample.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                throw new FileNotFoundException($"{_parameters.CertificateName} did not contain any certificate with a private key.");
+            }
+
+            Console.WriteLine($"Using certificate {certificate.Thumbprint} {certificate.Subject}");
+
+            return certificate;
+        }
+
+        private void ReadCertificatePassword()
```
azabbasi, line 122: Is there a reason we don't get this parameter as an argument in main?

azabbasi, line 122: I wonder if this is really necessary in the sample code. can't we just throw an error if user hasn't specified a password in the input parameters?

azabbasi (approved):

provisioning/Samples/device/X509Sample/ProvisioningDeviceClientSample.cs
drwill-ms: If I require it on the command-line, then the password will be visible to anyone watching. This function takes input but doesn't show your password.

The functionality existed before, and I decided to retain it. It seems reasonable to me.

Thoughts?

azabbasi: That sounds fair enough. We take connection strings as input parameters all over our sample codes also, one can argue the same logic should be applied elsewhere, but cert passwords are much easier to remember than a SAS token or a key that is much longer ...

drwill-ms: Good points.

PR #165, opened 2020-12-03
Revamp DPS TPM sample

drwill-ms:
I got TPM working on my laptop, so that is tested. I can't seem to get the EK for the simulator to know if it works, and provide the sample instructions on how to do that.


security/Samples/SecurityProviderTpmSimulator/SecurityProviderTpmSimulator.cs
```diff
@@ -65,17 +67,22 @@ public override byte[] Sign(byte[] data)
...
-        [SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Sample code resiliency")]
         public static void StopSimulatorProcess()
         {
-            foreach (var process in Process.GetProcessesByName(Path.GetFileNameWithoutExtension(s_simulatorExeName)))
+            if (s_simulatorProcessId != 0)
             {
                 try
                 {
-                    process?.Kill();
+                    var process = Process.GetProcessById(s_simulatorProcessId);
```
barustum, line 76: nit: it's clear by the method name that we are getting a Process by Id. But is it best practice to use Process instead of var?

barustum (approved):

PR #167, opened 2020-12-07
Update simulated device project

drwill-ms:
- Target .NET 5.0
- Take device connection string as a parameter
- Fix control-C handling
- Await async calls
- Specify message content encoding and type


iot-hub/Quickstarts/simulated-device/SimulatedDevice.cs
```diff
@@ -4,64 +4,97 @@
...
 
-            while (true)
+            while (!ct.IsCancellationRequested)
             {
                 double currentTemperature = minTemperature + rand.NextDouble() * 15;
                 double currentHumidity = minHumidity + rand.NextDouble() * 20;
 
                 // Create JSON message
-                var telemetryDataPoint = new
+                var messageBody = JsonSerializer.Serialize(
```
barustum, line 76: string

barustum (approved):

robinsh:
Just in case you didn’t know, there’s several of those “update simulated device project” projects and they’re all a little bit different. Are you going to make the same changes to all of then, or just as you come across them in the different projects?

drwill-ms:
> Just in case you didn’t know, there’s several of those “update simulated device project” projects and they’re all a little bit different. Are you going to make the same changes to all of then, or just as you come across them in the different projects?

I will, but I prefer to do them one at a time to get feedback as I go to avoid rework in subsequent projects and PRs that are too large.

robinsh:
I’m good with that. Just wanted to make sure I can rely on  you to take care of the ones you’re changing, so I know I don’t have to worry about it after the new year. BTW, Friday is my last day this year. )(Yikes!)

Robin

robinsh:
Awesome. Thanks David.

Robin

PR #169, opened 2020-12-09
Update DeviceSimulator2

drwill-ms:


iot-hub/Quickstarts/simulated-device-2/Program.cs
```diff
@@ -0,0 +1,147 @@
...
+        private static void CheckForConnectionStringArgument(string[] args)
+        {
+            if (args.Any())
+            {
+                try
+                {
+                    var cs = IotHubConnectionStringBuilder.Create(args.First());
+                    s_connectionString = cs.ToString();
+                }
+                catch (Exception) { }
```
azabbasi, line 71: if we are not doing anything in the catch, we probably don't need the try block either.
I personally think that if this throws we should exit with an error.

iot-hub/Quickstarts/simulated-device-2/Program.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        private static async Task Main(string[] args)
+        {
+            Console.WriteLine("IoT Hub Quickstarts #1 - Simulated device.");
+
+            // This sample accepts the device connection string as a parameter, if present
+            CheckForConnectionStringArgument(args);
+
+            // Connect to the IoT hub using the MQTT protocol
+            s_deviceClient = DeviceClient.CreateFromConnectionString(s_connectionString, TransportType.Mqtt);
```
azabbasi, line 41: suggestion: We can call out the TransportType as a private static field so it's easier to find/change by the users.

azabbasi (approved):
a couple of nit comments.

iot-hub/Quickstarts/simulated-device-2/Program.cs
drwill-ms: The parameter is optional, and I only want to set it if it passes parsing.

Issue #170, opened 2020-12-09
Quickstart device quick stream proxy is not working 


richardtans:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [ x] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
> cannot compile and build project in the quickstart
ServiceLocalProxyStreamingSample

### Any log messages given by the failure
>Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0246	The type or namespace name 'DeviceStreamRequest' could not be found (are you missing a using directive or an assembly reference?)	ServiceLocalProxyStreamingSample	C:\Users\Richard Lawidjaja\Desktop\1\azure-iot-samples-csharp-master\iot-hub\Quickstarts\device-streams-proxy\service\DeviceStreamSample.cs	54	Active

### Expected/desired behavior
>should be able to compiled

### OS and Version? win 10
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)

### Versions
>

### Mention any other details that might be useful

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

azabbasi:
Thank you for bringing this up to our attention.

PR #171, opened 2020-12-09
Fix QuickStart samples build

azabbasi:
We cannot use floating version in our DeviceStreaming samples because they need to have the preview versions selected and the floating version specification does not work for us because our preview versions are not considered the latest version in our history

image

As you can see here, 1.28.0 is considered a newer package than 1.28.0-preview-001 so we can't use floating version to fetch this package as per:
https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#floating-version-resolutions

### DIFF
--- a/iot-hub/Quickstarts/device-streams-proxy/device/Program.cs
+++ b/iot-hub/Quickstarts/device-streams-proxy/device/Program.cs
@@ -33,9 +33,9 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
         // Select one of the following transports used by DeviceClient to connect to IoT Hub.
         private static readonly TransportType s_transportType = TransportType.Amqp;
-        //private static TransportType s_transportType = TransportType.Mqtt;
-        //private static TransportType s_transportType = TransportType.Amqp_WebSocket_Only;
-        //private static TransportType s_transportType = TransportType.Mqtt_WebSocket_Only;
+        //private static readonly TransportType s_transportType = TransportType.Mqtt;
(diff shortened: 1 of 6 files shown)


azabbasi:
Fixes #170

iot-hub/Quickstarts/device-streams-proxy/device/Program.cs
```diff
@@ -32,7 +32,7 @@ public static class Program
         private static string s_port = Environment.GetEnvironmentVariable("REMOTE_PORT");
 
         // Select one of the following transports used by DeviceClient to connect to IoT Hub.
-        private static TransportType s_transportType = TransportType.Amqp;
+        private static readonly TransportType s_transportType = TransportType.Amqp;
         //private static TransportType s_transportType = TransportType.Mqtt;
```
barustum, line 36: update the rest to be readonly

azabbasi: we can't. Main() overrides them if they have been provided by the command line arguments.

barustum (approved):

timtay-microsoft (approved):

iot-hub/Quickstarts/device-streams-proxy/device/Program.cs
azabbasi: ah, you are talking about the commented code, I thought you meant the other static fields.
Fixed.

drwill-ms (approved):

PR #174, opened 2020-12-14
Add a step to install needed .NET SDKs to pipeline

barustum:


jamdavi (approved):

abhipsaMisra (approved):
what about netcore 3.1? is it available by default?

barustum:
> what about netcore 3.1? is it available by default?

Not sure. But based on the current run, it doesn't seem like it's needed:
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=74731&view=results

Will wait for the pipeline to finish before merging.

barustum:
> > what about netcore 3.1? is it available by default?
> 
> Not sure. But based on the current run, it doesn't seem like it's needed:
> https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=74731&view=results
> 
> Will wait for the pipeline to finish before merging.



> > what about netcore 3.1? is it available by default?
> 
> Not sure. But based on the current run, it doesn't seem like it's needed:
> https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=74731&view=results
> 
> Will wait for the pipeline to finish before merging.

Looks like it's needed. Adding it now.

drwill-ms (approved):

PR #178, opened 2021-02-03
Update FileUploadSample to use the new storage client

abhipsaMisra:
Turns out, customers directly copy over the code from samples to use in their applications.
For this reason, I am updating our FileUploadSample to use the non-deprecated storage client.


vinagesh (approved):

ChristianBeier:
please add a comment or outline that this sample will override the exiting blob. In most cases this is not a wished behaviour.

PR #181, opened 2021-02-10
Update device SDK version module sample

drwill-ms:


abhipsaMisra (commented):
Will this give a signal to customers that we are now officially supporting net5.0. 
While I agree that we should be on the latest and greatest version, we haven't incorporated it into our test pipeline yet. (edit: to be able to confidently say that our clients support it).

abhipsaMisra (approved):

bikamani (approved):

PR #182, opened 2021-02-12
AAD auth sample

vinagesh:
## Purpose

* ...

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

## Other Information
Sample to create token credentials using azure active directory and authenticate to IoT hub. I have added a sample that only calls the ServiceClient as I felt the main point here was to show how to construct the credentials.

### DIFF
--- a/iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
+++ b/iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
@@ -1,0 +1,57 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using CommandLine;
+using Microsoft.Azure.Devices;
+
+namespace RoleBasedAuthenticationSample
+{
+    /// <summary>
+    /// Parameters for the application.
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            'h',
+            "HostName",
+            Required = true,
+            HelpText = "The IoT hub host name. Ex: my-iot-hub.azure-devices.net")]
+
+        public string HostName { get; set; }
+
+        [Option(
+           'd',
+           "DeviceId",
+           Required = true,
+           HelpText = "The IoT hub device to send a message to.")]
+
+        public string DeviceId { get; set; }
+
+        [Option(
+            't',
+            "TransportType",
+            Default = TransportType.Amqp,
+            Required = false,
+            HelpText = "The transport to use to communicate with the IoT Hub. Possible values include Amqp and Amqp_WebSocket_Only.")]
+        
+        public TransportType TransportType { get; set; }
+
+        [Option(
+            "clientId",
+            Required = true,
+            HelpText = "The client Id of the Azure Active Directory application.")]
+        public string ClientId { get; set; }
+
+        [Option(
+            "tenantId",
+            Required = true,
+            HelpText = "The Azure Active Directory tenant (directory) Id ")]
+        public string TenantId { get; set; }
+
+        [Option(
+            "clientSecret",
+            Required = false,
+            HelpText = "A client secret that was generated for the application Registration used to authenticate the client. Only applicable when using LoginMethod of AppId.")]
+        public string ClientSecret { get; set; }
+    }
+}
--- a/iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
+++ b/iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Azure.Core;
+using Azure.Identity;
+using CommandLine;
+using Microsoft.Azure.Devices;
+using System;
+using System.Threading.Tasks;
+
+namespace RoleBasedAuthenticationSample
+{
+    public class Program
+    {
+        /// <summary>
+        /// A sample to illustrate how to use Azure active directory for authentication to the IoT hub.
+        /// <param name="args">Run with `--help` to see a list of required and optional parameters.</param>
+        public static async Task Main(string[] args)
+        {
+            // Parse application parameters
+            Parameters parameters = null;
+            ParserResult<Parameters> result = Parser.Default.ParseArguments<Parameters>(args)
+                .WithParsed(parsedParams =>
+                {
+                    parameters = parsedParams;
+                })
+                .WithNotParsed(errors =>
+                {
+                    Environment.Exit(1);
+                });
+
+            // Initialize Azure active directory credentials.
+            Console.WriteLine("Creating token credential.");
+
+            // These environment variables are necessary for DefaultAzureCredential to use application Id and client secret to login.
+            Environment.SetEnvironmentVariable("AZURE_CLIENT_SECRET", parameters.ClientSecret);
+            Environment.SetEnvironmentVariable("AZURE_CLIENT_ID", parameters.ClientId);
+            Environment.SetEnvironmentVariable("AZURE_TENANT_ID", parameters.TenantId);
+
+            // DefaultAzureCredential supports different authentication mechanisms and determines the appropriate credential type based of the environment it is executing in.
+            // It attempts to use multiple credential types in an order until it finds a working credential.
+            // For more info <see cref="https://docs.microsoft.com/en-us/dotnet/api/azure.identity?view=azure-dotnet"/>
+            TokenCredential tokenCredential = new DefaultAzureCredential(); 
+
+            // There are constructors for all the other clients where you can pass AAD credentials - JobClient, RegistryManager, DigitalTwinClient
+            using var serviceClient = ServiceClient.Create(parameters.HostName, tokenCredential, parameters.TransportType);
+
+            var sample = new RoleBasedAuthenticationSample();
+            await sample.RunSampleAsync(serviceClient, parameters.DeviceId);
+        }
+    }
+}
--- a/iot-hub/Samples/service/RoleBasedAuthenticationSample/RoleBasedAuthenticationSample.cs
+++ b/iot-hub/Samples/service/RoleBasedAuthenticationSample/RoleBasedAuthenticationSample.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Microsoft.Azure.Devices;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace RoleBasedAuthenticationSample
+{
+    /// <summary>
+    /// This sample connects to the IoT hub using Azure active directory credentials and sends a cloud-to-device message.
+    /// </summary>
+    public class RoleBasedAuthenticationSample
+    {
+        public async Task RunSampleAsync(ServiceClient client, string deviceId)
+        {
+            Console.WriteLine("Connecting using token credential.");
+            await client.OpenAsync();
+            Console.WriteLine("Successfully opened connection.");
+
+            Console.WriteLine("Sending a cloud-to-device message.");
+            using var message = new Message(Encoding.ASCII.GetBytes("Hello, Cloud!"));
+            await client.SendAsync(deviceId, message);
+            Console.WriteLine("Successfully sent message.");
+        }
+    }
+}
--- a/iot-hub/Samples/service/IoTHubServiceSamples.sln
+++ b/iot-hub/Samples/service/IoTHubServiceSamples.sln
@@ -31,6 +31,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TemperatureController", "Pn
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ColorConsoleLogger", "..\..\..\helpers\ColorConsoleLogger\ColorConsoleLogger.csproj", "{DBABF65A-F2E7-4736-9B99-D760E664B9C6}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "RoleBasedAuthenticationSample", "RoleBasedAuthenticationSample\RoleBasedAuthenticationSample.csproj", "{6629D377-B025-4798-BD30-18E47202AB75}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Azure.Devices", "..\..\..\..\azure-iot-sdk-csharp-track1\iothub\service\src\Microsoft.Azure.Devices.csproj", "{277FFAD5-18C4-427A-A305-4CE608DD249C}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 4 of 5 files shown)


iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+            // These environment variables are necessary for DefaultAzureCredential to use application Id and client secret to login.
+            Environment.SetEnvironmentVariable("AZURE_CLIENT_SECRET", parameters.ClientSecret);
+            Environment.SetEnvironmentVariable("AZURE_CLIENT_ID", parameters.ClientId);
+            Environment.SetEnvironmentVariable("AZURE_TENANT_ID", parameters.TenantId);
+
+            // DefaultAzureCredential supports different authentication mechanisms and determines the appropriate credential type based of the environment it is executing in.
+            // It attempts to use multiple credential types in an order until it finds a working credential.
+            // For more info <see cref="https://docs.microsoft.com/en-us/dotnet/api/azure.identity?view=azure-dotnet"/>
+            var tokenCredential = new DefaultAzureCredential(); 
+            var serviceClient = ServiceClient.Create(parameters.HostName, tokenCredential, parameters.TransportType);
```
vinagesh, line 42: I have chosen one of our clients to demo how to create a token credential and pass it in.

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+                    parameters = parsedParams;
+                })
+                .WithNotParsed(errors =>
+                {
+                    Environment.Exit(1);
+                });
+
+            // Initialize Azure active directory credentials.
+            Console.WriteLine("Creating token credential.");
+
```
vinagesh, line 32: This was the recommended way in track 2 to create Azure.Identity credential in SDK samples so following that pattern.

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+            // These environment variables are necessary for DefaultAzureCredential to use application Id and client secret to login.
+            Environment.SetEnvironmentVariable("AZURE_CLIENT_SECRET", parameters.ClientSecret);
+            Environment.SetEnvironmentVariable("AZURE_CLIENT_ID", parameters.ClientId);
+            Environment.SetEnvironmentVariable("AZURE_TENANT_ID", parameters.TenantId);
+
+            // DefaultAzureCredential supports different authentication mechanisms and determines the appropriate credential type based of the environment it is executing in.
+            // It attempts to use multiple credential types in an order until it finds a working credential.
+            // For more info <see cref="https://docs.microsoft.com/en-us/dotnet/api/azure.identity?view=azure-dotnet"/>
+            var tokenCredential = new DefaultAzureCredential();
```
timtay-microsoft, line 41: I'd consider making the ```var``` into ```TokenCredential``` so that it is more obvious that users don't have to use ```DefaultAzureCredential``` when constructing their service client

iot-hub/Samples/service/RoleBasedAuthenticationSample/RoleBasedAuthenticationSample.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            Console.WriteLine("Connecting using token credential.");
+            await client.OpenAsync();
+            Console.WriteLine("Success.");
+
+            Console.WriteLine("Sending a cloud-to-device message.");
+            var str = $"Hello, Cloud!";
+            var message = new Message(Encoding.ASCII.GetBytes(str));
+            await client.SendAsync(deviceId, message);
+            message.Dispose();
```
timtay-microsoft, line 26: Could you use a using statement here? Like:

```csharp
using var message = new Message(Encoding.ASCII.GetBytes(str));
await client.SendAsync(deviceId, message);
```

timtay-microsoft: It would spare you the message.Dispose() line

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
timtay-microsoft: I'm curious if Java has a similar pattern that it should follow. Where did you get this recommendation for .NET from?

timtay-microsoft:
>I have added a sample that only calls the ServiceClient as I felt the main point here was to show how to construct the credentials.

Do you think it is worth at least adding a note in this sample that the RegistryManager/DigitalTwinClient/etc. also have similar constructors, and that the same credential can be used when constructing them?

timtay-microsoft (approved):

iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <summary>
+    /// Parameters for the application.
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            'h',
+            "HostName",
+            Required = true,
+            HelpText = "The IoT hub host name.")]
```
abhipsaMisra, line 18: do you think an example would help here (my-iot-hub vs my-iot-hub.azure-devices.net)?

iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
```diff
@@ -0,0 +1,60 @@
...
+           "DeviceId",
+           Required = true,
+           HelpText = "The IoT hub device to send a message to.")]
+
+        public string DeviceId { get; set; }
+
+        [Option(
+            't',
+            "TransportType",
+            Default = TransportType.Amqp,
```
abhipsaMisra, line 33: we've been trying to get customers to default to mqtt instead of amqp (with mqtt becoming a bigger focus these days), we should probably do the same here as well

abhipsaMisra, line 33: discard my previous comment, this is service client, so amqp only!

iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
```diff
@@ -0,0 +1,60 @@
...
+            Required = false,
+            HelpText = "The transport to use to communicate with the IoT Hub. Possible values include Amqp and Amqp_WebSocket_Only.")]
+        
+        public TransportType TransportType { get; set; }
+
+        [Option(
+            'i',
+            "clientId",
+            Required = true,
+            HelpText = "Client Id of the application Id to login, or the application Id used to log the user in.")]
```
abhipsaMisra, line 43: I am a bit confused as to what this statement means - "Client Id of the application Id"

iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
```diff
@@ -0,0 +1,60 @@
...
+        public string ClientId { get; set; }
+
+        [Option(
+            't',
+            "tenantId",
+            Required = true,
+            HelpText = "Application tenant Id")]
```
abhipsaMisra, line 50: Is this the AAD app's tenant ID? should we be a bit more descriptive?

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
abhipsaMisra, line 42: using var serviceClient ...

abhipsaMisra (approved):

iot-hub/Samples/service/RoleBasedAuthenticationSample/RoleBasedAuthenticationSample.cs
vinagesh: yes thanks :) copy paste somehow blinds me at times :P

iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        [Option(
+            'i',
+            "clientId",
+            Required = true,
+            HelpText = "Client Id of the application Id to login, or the application Id used to log the user in.")]
+        public string ClientId { get; set; }
+
+        [Option(
+            't',
```
drwill-ms, line 47: `t` is already the short name for transport

iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
vinagesh: yes it will

vinagesh:
> > I have added a sample that only calls the ServiceClient as I felt the main point here was to show how to construct the credentials.
> 
> Do you think it is worth at least adding a note in this sample that the RegistryManager/DigitalTwinClient/etc. also have similar constructors, and that the same credential can be used when constructing them?

Good point. will include.

iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
vinagesh: oops! probably best to skip shortname here.

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
drwill-ms: One can only specify the values via environment variables, so given we take them as parameters, this is the only way for us to specify them.

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
drwill-ms: What? I don't get this.

iot-hub/Samples/service/RoleBasedAuthenticationSample/RoleBasedAuthenticationSample.cs
```diff
@@ -0,0 +1,32 @@
...
+            Console.WriteLine("Success.");
+
+            await client.CloseAsync();
```
drwill-ms, line 29: I don't think we like to promote calling `CloseAsync()`. @abhipsaMisra ?

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
drwill-ms: We don't do that in track2. This is how it was suggested by the Azure SDK team.

drwill-ms: At most, I'd put a link to all the different kinds of TokenCredential types and options.

drwill-ms (approved):
Lots of good suggestions posted by others that we should take.

iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
vinagesh: It should probably be application. I took it from track 2. Looks like we need an update over there too. Let me find better descriptions for these.

iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
vinagesh: yes it is. Same as above. Let me find better description for all the AAD stuff and add.

iot-hub/Samples/service/RoleBasedAuthenticationSample/RoleBasedAuthenticationSample.cs
vinagesh: Sure but why not?

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Azure.Devices;
+using System;
+using System.Threading.Tasks;
+
+namespace RoleBasedAuthenticationSample
+{
+    public class Program
+    {
+        /// <summary>
+        /// A sample to illustrate how to use Azure active directory for authentication to the IoT hub.
```
bikamani, line 14: will you be adding readme for this sample?

iot-hub/Samples/service/RoleBasedAuthenticationSample/RoleBasedAuthenticationSample.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <summary>
+    /// This sample connects to the IoT hub using Azure active directory credentials and sends a cloud-to-device message.
+    /// </summary>
+    public class RoleBasedAuthenticationSample
+    {
+        public async Task RunSampleAsync(ServiceClient client, string deviceId)
+        {
+            Console.WriteLine("Connecting using token credential.");
+            await client.OpenAsync();
+            Console.WriteLine("Success.");
```
bikamani, line 20: nit: good to restate logging as Success opening connection or something of that sort

iot-hub/Samples/service/RoleBasedAuthenticationSample/Parameters.cs
bikamani: Client id of App registration?

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
vinagesh: I got this recommendation from the SDK team when we were writing this for ADT. It might be there for Java as well. Would be good to check. The auth folks want us to use DefaultAzureCredential. And if users wants other things, they want them to refer to Azure.Identiy.

iot-hub/Samples/service/RoleBasedAuthenticationSample/RoleBasedAuthenticationSample.cs
abhipsaMisra: On the device client side, yeah, `CloseAsync()` isn't really useful right now since we cannot guarantee that the transport layer is cleaned up properly. For the service client, since our clients do not come with any recoverable retry logic anyway, we recommend that they dispose and recreate the client when they see an exception that tells then that the transport layer (or amqp links) was disconnected. 

I think what we want here is to avoid people following the open -> close -> open flow, so leaving it out of our samples and instead using dispose to control the lifecycle will help promote that.

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
vinagesh: Can you explain? Do we keep some readme somewhere?

vinagesh:
> Lots of good suggestions posted by others that we should take.

Already taken and updated.

iot-hub/Samples/service/RoleBasedAuthenticationSample/RoleBasedAuthenticationSample.cs
abhipsaMisra: Nothing will go wrong with the flow that we have now for this sample -> calling close at the end of this sample, but it could possibly encourage people to use it in their applications and run into inconsistent state issues.

vinagesh: I can remove it then. Thanks.

iot-hub/Samples/service/RoleBasedAuthenticationSample/Program.cs
vinagesh: Chatted offline. Will send a follow up PR with readme.

PR #183, opened 2021-02-12
Adding sample for SAS credential

vinagesh:
## Purpose

* ...

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Parameters.cs
+++ b/iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Parameters.cs
@@ -1,0 +1,55 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using CommandLine;
+using Microsoft.Azure.Devices;
+
+namespace IotHubSasCredentialAuthenticationSample
+{
+    /// <summary>
+    /// Parameters for the application.
+    /// </summary>
+    internal class Parameters
+    {
+        /// <summary>
+        /// The resource that the shared access token should grant access to. For cases where the token will be used for 
+        /// more than one function(i.e.used by registryManager to create a device and used by serviceClient to send cloud 
+        /// to device messages), this value should be the hostName of your IoT Hub ("my-azure-iot-hub.azure-devices.net"
+        /// for example). Shared access signatures do support scoping of the resource authorization by making this resourceUri
+        /// more specific.For example, a resourceUri of "my-azure-iot-hub.azure-devices.net/devices" will make this token only usable
+        /// when creating/updating/deleting device identities. 
+        /// For more examples, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#use-security-tokens-from-service-components"/>
+        /// </summary>
+        [Option(
+            'r',
+            "ResourceUri",
+            Required = true,
+            HelpText = "The resource that the shared access token should grant access to. Ex: my-iot-hub.azure-devices.net")]
+
+        public string ResourceUri { get; set; }
+
+        [Option(
+           'd',
+           "DeviceId",
+           Required = true,
+           HelpText = "The IoT hub device to send a message to.")]
+
+        public string DeviceId { get; set; }
+
+        [Option(
+            't',
+            "TransportType",
+            Default = TransportType.Amqp,
+            Required = false,
+            HelpText = "The transport to use to communicate with the IoT Hub. Possible values include Amqp and Amqp_WebSocket_Only.")]
+        
+        public TransportType TransportType { get; set; }
+
+        [Option(
+            's',
+            "SharedAccessKey",
+            Required = true,
+            HelpText = "The shared access key for IoT hub.")]
+        public string SharedAccessKey { get; set; }       
+    }
+}
--- a/iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
+++ b/iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
@@ -1,0 +1,87 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Azure;
+using Azure.Core;
+using CommandLine;
+using Microsoft.Azure.Devices;
+using System;
+using System.Globalization;
+using System.Net;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace IotHubSasCredentialAuthenticationSample
+{
+    public class Program
+    {
+        /// <summary>
+        /// A sample to illustrate how to use SAS token for authentication to the IoT hub.
+        /// <param name="args">Run with `--help` to see a list of required and optional parameters.</param>
+        public static async Task Main(string[] args)
+        {            
+            // Parse application parameters
+            Parameters parameters = null;
+            ParserResult<Parameters> result = Parser.Default.ParseArguments<Parameters>(args)
+                .WithParsed(parsedParams =>
+                {
+                    parameters = parsedParams;
+                })
+                .WithNotParsed(errors =>
+                {
+                    Environment.Exit(1);
+                });
+
+            // Initialize SAS token credentials.
+            Console.WriteLine("Creating sas credential.");
+
+            TimeSpan timeToLive = TimeSpan.FromHours(1);
+            DateTime expiresOn = DateTime.UtcNow.Add(timeToLive);
+            string sasToken = GenerateSasToken(parameters.ResourceUri, parameters.SharedAccessKey, expiresOn);
+            // Note: Pass the generated sasToken and not just the shared access signature when creating the AzureSasCredential.
+            AzureSasCredential sasCredential = new AzureSasCredential(sasToken);
+
+            // This is how the credential can be updated in the AzureSasCredential object whenever necessary.
+            // This sample just shows how to perform the update but it is not necessary to update the token
+            // until the token is close to its expiry.
+            DateTime newExpiresOn = DateTime.UtcNow.Add(timeToLive);
+            string updatedSasToken = GenerateSasToken(parameters.ResourceUri, parameters.SharedAccessKey, newExpiresOn);
+            sasCredential.Update(updatedSasToken);
+
+            // There are constructors for all the other clients where you can pass SAS credentials - JobClient, RegistryManager, DigitalTwinClient
+            var hostName = parameters.ResourceUri.Split('/')[0];
+            using var serviceClient = ServiceClient.Create(hostName, sasCredential, parameters.TransportType);
+
+            var sample = new IotHubSasCredentialAuthenticationSample();
+            await sample.RunSampleAsync(serviceClient, parameters.DeviceId);
+        }
+
+        private static string GenerateSasToken(string resourceUri, string sharedAccessKey, DateTime expiresOn, string policyName = default )
+        {
+            DateTime epochTime = new DateTime(1970, 1, 1);
+            TimeSpan secondsFromEpochTime = expiresOn.Subtract(epochTime);
+            long seconds = Convert.ToInt64(secondsFromEpochTime.TotalSeconds, CultureInfo.InvariantCulture);
+            string expiry = Convert.ToString(seconds, CultureInfo.InvariantCulture);
+
+            string stringToSign = WebUtility.UrlEncode(resourceUri) + "\n" + expiry;
+
+            HMACSHA256 hmac = new HMACSHA256(Convert.FromBase64String(sharedAccessKey));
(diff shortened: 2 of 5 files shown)


iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,86 @@
...
+            TimeSpan timeToLive = TimeSpan.FromHours(1);
+            DateTime expiresOn = DateTime.UtcNow.Add(timeToLive);
+            IotHubSasCredential sasCredential = new IotHubSasCredential(
+                GenerateSasToken(parameters.HostName, parameters.SharedAccessKey, expiresOn),
+                expiresOn);
+
+            // This is how the credential can be updated in the IotHubSasCredential object whenever necessary. It is not 
+            // necessary to do before the token expires. The following just demonstrates how to do it.
+            // NOTE: Ensure to pass the new expiry.
+            DateTime newExpiresOn = DateTime.UtcNow.Add(timeToLive);
```
vinagesh, line 46: As we have our own derived class, that class now exposes two overloads for Update. And if the user uses the one without the expiryDate parameter, we will be passing he wrong date downstream. This means we have to add some code to disable that method. The only way to do it is by throwing in a override as in this example draft PR - https://github.com/Azure/azure-iot-sdk-csharp/pull/1793/files

This made me re-think our approach here and want to know other opinions. I am not completely against the approach in the above PR as it keeps us from doing any parsing of the SAS token. But it does mean that there can be a small trial and error learning curve in case users don't read our documentation.

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,86 @@
...
+            sasCredential.Update(
+                GenerateSasToken(parameters.HostName, parameters.SharedAccessKey, newExpiresOn),
+                newExpiresOn);
+
+            // There are constructors for all the other clients where you can pass AAD credentials - JobClient, RegistryManager, DigitalTwinClient
```
bikamani, line 51: Nit: AAD --> Sas

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
abhipsaMisra: Instead of taking in the DateTime in the constructor, is taking a ttl an option? That way we could use the ttl supplied in the constructor in case they don't provide one for the Update method.

abhipsaMisra: We could use the ttl to generate the DateTime, if that is what we and possibly the base credential class require.

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace IotHubSasCredentialAuthenticationSample
+{
+    public class Program
+    {
+        /// <summary>
+        /// A sample to illustrate how to use Azure active directory for authentication to the IoT hub.
```
abhipsaMisra, line 18: typo: sas here

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
vinagesh: What if they change the ttl? Though chances are rare, I don't think we should make an assumption on that.

vinagesh: The AMQP layer expects us to send the DateTime of expiry. So the initial design was to get it from user instead of parsing the token.

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,86 @@
...
+            Console.WriteLine("Creating sas credential.");
+
+            TimeSpan timeToLive = TimeSpan.FromHours(1);
+            DateTime expiresOn = DateTime.UtcNow.Add(timeToLive);
+            IotHubSasCredential sasCredential = new IotHubSasCredential(
+                GenerateSasToken(parameters.HostName, parameters.SharedAccessKey, expiresOn),
+                expiresOn);
+
+            // This is how the credential can be updated in the IotHubSasCredential object whenever necessary. It is not 
+            // necessary to do before the token expires. The following just demonstrates how to do it.
```
bikamani, line 44: would you like to understand what you meant to say by "It is not necessary to do before the token expires."

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
vinagesh: And generating the DateTime can have some clock skew (it might be acceptable but it will happen). It will not be the same as the one on the SAS token.

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
vinagesh: sure. The SAS token expires. So update need not be called unnecessarily. In the sample, I'm calling it as soon as I created a token just to show the usage of the method.

vinagesh: I can improve the wordings a bit.

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,87 @@
...
+            // Initialize SAS token credentials.
+            Console.WriteLine("Creating sas credential.");
+
+            TimeSpan timeToLive = TimeSpan.FromHours(1);
+            DateTime expiresOn = DateTime.UtcNow.Add(timeToLive);
+            IotHubSasCredential sasCredential = new IotHubSasCredential(
+                GenerateSasToken(parameters.HostName, parameters.SharedAccessKey, expiresOn),
+                expiresOn);
+
+            // This is how the credential can be updated in the IotHubSasCredential object whenever necessary.
```
vinagesh, line 43: @bikamani - Is this improved wording more clear?

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
bikamani: How will ttl provided at constructor time be valid/used for update operation?

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
bikamani: Very precisely now, yes. 👍

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
vinagesh: This is now fixed by using encapsulation instead of inheritance. Only one update method will be exposed now. See PR - https://github.com/Azure/azure-iot-sdk-csharp/pull/1793

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,87 @@
...
+            sasCredential.Update(updatedSasToken);
+
+            // There are constructors for all the other clients where you can pass SAS credentials - JobClient, RegistryManager, DigitalTwinClient
+            using var serviceClient = ServiceClient.Create(parameters.HostName, sasCredential, parameters.TransportType);
+
+            var sample = new IotHubSasCredentialAuthenticationSample();
+            await sample.RunSampleAsync(serviceClient, parameters.DeviceId);
+        }
+
+        private static string GenerateSasToken(string hostName, string sharedAccessKey, DateTime expiresOn, string policyName = default )
```
timtay-microsoft, line 59: In the Java sample, I made this function allow for security scoping. Instead of asking for the hostname, I asked for the resource URI that the SAS token should grant access to.

timtay-microsoft:
```java
@param resourceUri The resource that the shared access token should grant access to. For cases where the token
 * will be used for more than one function (i.e. used by registryManager to create a device and used by serviceClient
 * to send cloud to device messages), this value should be the hostName of your IoT Hub
 * ("my-azure-iot-hub.azure-devices.net" for example). Shared access signatures do support scoping of the resource
 * authorization by making this resourceUri more specific. For example, a resourceUri of "my-azure-iot-hub.azure-devices.net/devices"
 * will make this token only usable when creating/updating/deleting device identities.
 * For more examples, see <a href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#use-security-tokens-from-service-components">Using security tokens from service components</a>
```

timtay-microsoft: I think a customer who is interested in generating SAS tokens would be interested in this feature, so consider leaving a note about this scoping option

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            // SharedAccessSignature sr=ENCODED(dh://myiothub.azure-devices.net/a/b/c?myvalue1=a)&sig=<Signature>&se=<ExpiresOnValue>[&skn=<KeyName>]
+            string token = string.Format(
+                CultureInfo.InvariantCulture,
+                "SharedAccessSignature sr={0}&sig={1}&se={2}",
+                WebUtility.UrlEncode(hostName),
+                WebUtility.UrlEncode(signature),
+                expiry);
+
+            if (!string.IsNullOrWhiteSpace(policyName))
```
timtay-microsoft, line 79: Good idea making the policy name an optional parameter. I need to do the same on the Java sample code.

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Parameters.cs
```diff
@@ -0,0 +1,46 @@
...
+            "TransportType",
+            Default = TransportType.Amqp,
+            Required = false,
+            HelpText = "The transport to use to communicate with the IoT Hub. Possible values include Amqp and Amqp_WebSocket_Only.")]
+        
+        public TransportType TransportType { get; set; }
+
+        [Option(
+            's',
+            "ShareAccessKey",
```
timtay-microsoft, line 41: "Share**d**AccessKey"

timtay-microsoft (approved):

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,87 @@
...
+            TimeSpan timeToLive = TimeSpan.FromHours(1);
+            DateTime expiresOn = DateTime.UtcNow.Add(timeToLive);
+            string sasToken = GenerateSasToken(parameters.HostName, parameters.SharedAccessKey, expiresOn);
+            // Note: Pass the generated sasToken and not just the shared access signature when creating the AzureSasCredential.
+            AzureSasCredential sasCredential = new AzureSasCredential(sasToken);
+
+            // This is how the credential can be updated in the IotHubSasCredential object whenever necessary.
+            // This sample just shows how to perform the update but it is not necessary to update the token
+            // until the token is close to its expiry.
+            // NOTE: Ensure to pass the new expiry date time.
```
abhipsaMisra, line 47: nit: leftover comment from previous impl :)

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
```diff
@@ -0,0 +1,87 @@
...
+            DateTime newExpiresOn = DateTime.UtcNow.Add(timeToLive);
+            string updatedSasToken = GenerateSasToken(parameters.HostName, parameters.SharedAccessKey, expiresOn);
```
abhipsaMisra, line 49: `newExpiresOn` to be passed in, instead of `expiresOn`

abhipsaMisra (approved):

iot-hub/Samples/service/IotHubSasCredentialAuthenticationSample/Program.cs
vinagesh: Upadated

PR #185, opened 2021-02-25
Update the device sample to use callbacks for C2D messaging

abhipsaMisra:


iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -68,7 +70,15 @@ public async Task RunSampleAsync(TimeSpan sampleRunningTime)
             try
             {
                 await InitializeAndOpenClientAsync();
-                await Task.WhenAll(SendMessagesAsync(cts.Token), ReceiveMessagesAsync(cts.Token));
+
+                if (_transportType.Equals(TransportType.Http1))
```
drwill-ms, line 74: Why `.Equals()` and not `==`?

abhipsaMisra: I tend to associate == with comparison by reference, while .Equals() as comparison by value; so I chose to use .Equals() here

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -106,6 +119,21 @@ private async Task InitializeAndOpenClientAsync()
...
                     _logger.LogDebug($"Opened the client instance.");
+
+                    switch (_transportType)
+                    {
+                        case TransportType.Amqp:
+                        case TransportType.Amqp_Tcp_Only:
+                        case TransportType.Amqp_WebSocket_Only:
+                        case TransportType.Mqtt:
+                        case TransportType.Mqtt_Tcp_Only:
+                        case TransportType.Mqtt_WebSocket_Only:
```
abhipsaMisra, line 130: this could also have been if (_transportType is not Http1) -> subscribe to c2d; but I wanted to call out that all these transport protocols supports c2d receive callback; so I went with the switch case

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -68,7 +70,15 @@ public async Task RunSampleAsync(TimeSpan sampleRunningTime)
             try
             {
                 await InitializeAndOpenClientAsync();
-                await Task.WhenAll(SendMessagesAsync(cts.Token), ReceiveMessagesAsync(cts.Token));
+
+                if (_transportType.Equals(TransportType.Http1))
+                {
+                    await TaskHelper.WhenAllFailFast(SendMessagesAsync(cts.Token), ReceiveMessagesAsync(cts.Token), CompleteReceivedMessagesAsync(cts.Token));
```
abhipsaMisra, line 76: Tas.WhenAll() will wait until all supplied tasks complete before returning the exception, but I wanted the application to fail as soon as a non-recoverable exception was encountered; hence this fail fast implementation

abhipsaMisra: his is esp because the supplied tasks are long-running which will terminate only when cancelled explicitly or if a non-recoverable exception is encountered

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -97,6 +107,9 @@ private async Task InitializeAndOpenClientAsync()
 
                     s_deviceClient = DeviceClient.CreateFromConnectionString(_deviceConnectionStrings.First(), _transportType, _clientOptions);
                     s_deviceClient.SetConnectionStatusChangesHandler(ConnectionStatusChangeHandler);
+
+                    // Updating the client's operation timeout to 30secs for the purpose of a sample. The default value is 4mins.
+                    s_deviceClient.OperationTimeoutInMilliseconds = (uint)s_deviceOperationTimeout.TotalMilliseconds;
```
abhipsaMisra, line 112: this is to encourage a faster timeout in case a persistent retryable exception is encountered

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -210,16 +238,15 @@ private async Task SendMessagesAsync(CancellationToken cancellationToken)
...
-                        {
-                            _logger.LogError($"A network related exception was caught, but will try to recover and retry: {ex}");
+                        catch (UnauthorizedException ex)
+                        {
+                            _logger.LogError($"An UnauthorizedException was caught, recovery will be handled by the ConnectionStatusChangeHandler: {ex}");
                         }
-                        catch (Exception ex)
+                        catch (Exception ex) when (ExceptionHelper.IsNetworkExceptionChain(ex))
                         {
-                            _logger.LogError($"Unexpected error {ex}");
```
abhipsaMisra, line 222: no longer consuming a non-recoverable exception, instead causing the application to terminate

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms, line 76: Some amount of DRY violation here. You could create a list and initialize it with the 2 common tasks, then optionally add the `ReceiveMessagesAsync` task inside an if statement. Then after, use TaskHelper.WhenAllFailFast on the list.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: Transport type is an enum, effectively an int, so that is irrelevant in this case, right?

I'm pretty sure in C# the default should be == unless you explicitly need `Equals()`.

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
```diff
@@ -78,6 +78,7 @@ await deviceClient.CompleteAsync(receivedMessage);
 ```
 
 ### Receive cloud to device telemetry (using the callback) and complete the message:
```
drwill-ms, line 80: always have empty lines above and below headers

iot-hub/Samples/device/DeviceReconnectionSample/readme.md
```diff
@@ -78,6 +78,7 @@ await deviceClient.CompleteAsync(receivedMessage);
...
+* NOTE: This API is available only on MQTT and AMQP.
```
drwill-ms, line 81:
```
> Note: ...
```

drwill-ms: Should look like:

> Note: blah blah blah

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -210,16 +238,15 @@ private async Task SendMessagesAsync(CancellationToken cancellationToken)
...
-                            // Inspect the exception to figure out if operation should be retried, or if user-input is required.
-                            _logger.LogError($"An IotHubException was caught, but will try to recover and retry: {ex}");
+                            _logger.LogError($"A transient IotHubException was caught, will retry sending the message: {ex}");
                         }
-                        catch (Exception ex) when (ExceptionHelper.IsNetworkExceptionChain(ex))
-                        {
-                            _logger.LogError($"A network related exception was caught, but will try to recover and retry: {ex}");
+                        catch (UnauthorizedException ex)
+                        {
+                            _logger.LogError($"An UnauthorizedException was caught, recovery will be handled by the ConnectionStatusChangeHandler: {ex}");
```
drwill-ms, line 245: In changing this log, the part after the comma is now an independent sentence. That means you should replace the comma with a semicolon or a period.

drwill-ms: Or you could insert "and"

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -210,16 +238,15 @@ private async Task SendMessagesAsync(CancellationToken cancellationToken)
...
                         }
-                        catch (Exception ex)
+                        catch (Exception ex) when (ExceptionHelper.IsNetworkExceptionChain(ex))
                         {
-                            _logger.LogError($"Unexpected error {ex}");
+                            _logger.LogError($"A network related exception was caught, will retry sending the message: {ex}");
```
drwill-ms, line 249: Same here.

Also you need to be more explicit about the subject of the second clause. Who will retry? The device app or the SDK?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -231,37 +258,102 @@ private async Task SendMessagesAsync(CancellationToken cancellationToken)
...
+
+            _receivedMessagesQueue.Enqueue(receivedMessage);
+            return Task.CompletedTask;
+        }
+
+        private async Task CompleteReceivedMessagesAsync(CancellationToken cancellationToken)
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                if (IsDeviceConnected)
```
drwill-ms, line 315: To prevent such massive indenting, how about handling error conditions different?

Like:

```csharp
if (!IsDeviceConnected)
{
    continue;
}
```

That way the happy path code doesn't need to be indented quite so many times?

You'd probably need to move the Task.Delay to the top of the loop though, then.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -231,37 +258,102 @@ private async Task SendMessagesAsync(CancellationToken cancellationToken)
...
+                {
+                    if (_receivedMessagesQueue.TryPeek(out Message messageToBeCompleted))
```
drwill-ms, line 317: Why use TryPeek instead of TryDequeue?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: Why not `Task.WhenAny()`?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
vinagesh: Adding a case for Http1 and returning "OperationNotAlllowed" exception would make this switch case better.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -231,37 +258,102 @@ private async Task SendMessagesAsync(CancellationToken cancellationToken)
...
+                    {
+                        using Message receivedMessage = await s_deviceClient.ReceiveAsync();
+                        if (receivedMessage == null)
+                        {
+                            _logger.LogInformation("No message received; timed out.");
+                            await Task.Delay(s_sleepDuration);
+                            continue;
+                        }
+
+                        await CloudToDeviceMessageHandler(receivedMessage, null);
```
vinagesh, line 280: Should the while loop be terminated after this?

abhipsaMisra:
This has been put on hold for now, closing this PR

PR #187, opened 2021-03-23
Update Program.cs

Rickyp:
Include the millisecond portion when printing the date

## Purpose

* The code was previously printing the DateTime of the enqueuedTime message system property without the millisecond portion, giving users the idea that iot hub was dropping the milliseconds somewhere before routing to the endpoint, which is not the case. The reason is because by default printing DateTime as is will not include the milliseconds as explained [here](https://docs.microsoft.com/en-us/dotnet/standard/base-types/how-to-display-milliseconds-in-date-and-time-values). This ensures that if printing a DateTime object, it will do it in the format required to include the millisecond portion.

```
[ ] Yes
[X ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe:
``` Addressing the confusion that iot hub is not dropping the milliseconds from the enqueuedTime system property.

## How to Test
*  Run the ReadD2cMessages project, iothub-enqueuedtime property should now also include the millisecond portion.

## Other Information
<!-- Add any other helpful information that may be needed here. -->

### DIFF
--- a/iot-hub/Quickstarts/ReadD2cMessages/Program.cs
+++ b/iot-hub/Quickstarts/ReadD2cMessages/Program.cs
@@ -100,13 +100,13 @@ namespace ReadD2cMessages
                     Console.WriteLine("\tApplication properties (set by device):");
                     foreach (KeyValuePair<string, object> prop in partitionEvent.Data.Properties)
                     {
-                        Console.WriteLine($"\t\t{prop.Key}: {prop.Value}");
+                        PrintProperties(prop);
                     }
 
                     Console.WriteLine("\tSystem properties (set by IoT Hub):");
                     foreach (KeyValuePair<string, object> prop in partitionEvent.Data.SystemProperties)
                     {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: iot-hub/Quickstarts/ReadD2cMessages/Program.cs [hunk 1]
```diff
@@ -116,5 +116,17 @@ await foreach (PartitionEvent partitionEvent in consumer.ReadEventsAsync(ct))
                 // error in this scenario.
             }
         }
+
+        private static void PrintProperties(KeyValuePair<string, object> prop) 
+        {
+            string propValue = prop.Value.ToString();
```
drwill-ms, line 122: Avoid converting the property to a string the first time when it is a DateTime.

```csharp
string propValue = prop.Value is DateTime
    ? ((DateTime)prop.Value).ToString("O") // using a built-in date format here that includes milliseconds
    : prop.Value.ToString();
```

drwill-ms (approved):

PR #188, opened 2021-03-30
Update device cleanup samples to use import/export jobs

abhipsaMisra:
In this PR:
- Update the devices cleanup sample to use import/export jobs.
- Update device samples to use latest Microsoft.Azure.Devices.Client nuget
- Update method and twin sample to accept sample running time as optional input (these samples have callbacks set that were previously waiting for 5 mins for user input. While waiting for 5mins does give a good sample experience, it would also slow down our samples pipeline run. The sample running time is now configurable. (sample cancellation using ctrl+c is still supported).
- Update file upload sample to force-upload blobs.


build.ps1
```diff
@@ -108,7 +108,7 @@ try {
 
         # Run cleanup first so the samples don't get overloaded with old devices
         RunApp provisioning\Samples\service\CleanupEnrollmentsSample "Provisioning\Service\CleanupEnrollmentsSample"
-        RunApp iot-hub\Samples\service\CleanUpDevicesSample "IoTHub\Service\CleanUpDevicesSample"
+        RunApp iot-hub\Samples\service\CleanupDevicesSample "IoTHub\Service\CleanupDevicesSample" "-i ""$env:IOTHUB_CONNECTION_STRING"" -s ""$env:STORAGE_ACCOUNT_CONNECTION_STRING"" -p ""$env:PATH_TO_DEVICE_PREFIX_FOR_DELETION_FILE"""
```
drwill-ms, line 111: Passing environment variable values as parameters?

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,168 @@
...
+using Azure.Storage.Sas;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Samples
+{
+    public class CleanupDevicesSample
```
drwill-ms, line 15: Whole class could use more commenting as to approach.

iot-hub/Samples/service/CleanupDevicesSample/ImportExportDevicesHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        /// <summary>
+        /// Creates an enumerable of <see cref="ExportImportDevice"/> from the exported <see cref="Stream"/> data.
+        /// </summary>
+        /// <param name="inputStream">The exported <see cref="Stream"/> data.</param>
+        internal static IEnumerable<ExportImportDevice> BuildExportImportDeviceFromStream(Stream inputStream)
+        {
+            var exportedDevices = new List<ExportImportDevice>();
+
+            string content = new StreamReader(inputStream).ReadToEnd();
```
drwill-ms, line 45: `StreamReader` is `IDisposable` and should be disposed here.

iot-hub/Samples/service/CleanupDevicesSample/Parameters.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Microsoft.Azure.Devices.Client.Samples
+{
+    /// <summary>
+    /// Parameters for the application.
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            'i',
```
drwill-ms, line 14: In other samples, we used 'p' for primary connection string.

iot-hub/Samples/service/CleanupDevicesSample/Parameters.cs
```diff
@@ -0,0 +1,34 @@
...
+            "IotHubConnectionString",
+            Required = true,
+            HelpText = "The primary connection string for the IoT Hub instance to connect to.")]
+        public string IotHubConnectionString { get; set; }
+
+        [Option(
+            's',
```
drwill-ms, line 21: In other samples, we used 's' for secondary hub connection string. Although we don't use that here, it is a potential conflict to be aware of.

I think it is fine to use 's' here but I wanted you to be aware.

iot-hub/Samples/service/CleanupDevicesSample/ImportExportDevicesHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            var devicesFileSb = new StringBuilder();
+
+            foreach (ExportImportDevice device in devices)
+            {
+                devicesFileSb.AppendLine(JsonConvert.SerializeObject(device));
+            }
+
+            byte[] devicesFileInBytes = Encoding.Default.GetBytes(devicesFileSb.ToString());
+            return new MemoryStream(devicesFileInBytes);
```
drwill-ms, line 34: `MemoryStream` is IDisposable. I can't find the code that calls this method, but I wanted to make sure you checked that it was being disposed by the caller.

build.ps1
abhipsaMisra: yes, these values are available to the pipeline as environment variables (from keyvault): https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/vsts/vsts.yaml#L47

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
abhipsaMisra: I've added a comment for each step taken. Is there some other information that you'd like to see here? (maybe about specific access permissions granted to the sas token?)

iot-hub/Samples/service/CleanupDevicesSample/ImportExportDevicesHelper.cs
abhipsaMisra: It wasn't being disposed, I'll fix it.

iot-hub/Samples/service/CleanupDevicesSample/ImportExportDevicesHelper.cs
abhipsaMisra: will do

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,168 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
bikamani, line 1: since you are already refactoring, we can add some readme to this folder explaining about this sample.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Samples
+{
+    public class CleanupDevicesSample
+    {
+        private const string ImportExportDevicesFileName = "devices.txt";
```
bikamani, line 17: Is this file created as part of exporting device identities?

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
abhipsaMisra: There isn't much to explain here, as in, it isn't a long running sample that follows a certain convention. I've marked out the individual steps in the sample itself, but if there is something additional you'd like me to add (or a step that isn't clear to you), let me know and I'll add it in.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
abhipsaMisra: Yes, it is.
I'll add a link to the doc that describes the process, in our sample: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-bulk-identity-mgmt

bikamani: Cool, this is where I think readme would be useful. Even adding a link to this document and explaining the steps you have in comments would be awesome.

That also brings another question - Do the snippets in this doc need to be updated as well?

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        private static readonly TimeSpan _waitDuration = TimeSpan.FromSeconds(5);
```
bikamani, line 19: s_waitDuration

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
abhipsaMisra: I've added a link to the doc in the sample.

As for the doc snippets - they've used v11 of the storage sdk, and the snippets in the doc refer to the types available in there. (the ones I used in the sample are from v12). It would be nice to have the doc updated to the latest version of storage sdk, but I am not sure if that is something the owning team wants to do.
If this is something you're interested in and would like to pursue, that'd be great. It would definitely benefit customers to have code snippets that are compatible with the latest version of the sdk.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,168 @@
...
+        private static readonly IReadOnlyList<JobStatus> s_completedJobs = new[]
+        {
+            JobStatus.Completed,
+            JobStatus.Failed,
+            JobStatus.Cancelled,
+        };
+
+        private readonly RegistryManager _registryManager;
+        private readonly BlobContainerClient _blobContainerClient;
+        private readonly List<string> _deleteDeviceWithPrefix;
```
bikamani, line 29: _deleteDevice**s**WithPrefix

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,168 @@
...
+        private async Task PrintDeviceCountAsync()
+        {
+            string countSqlQuery = "SELECT COUNT() AS numberOfDevices FROM devices";
+            IQuery countQuery = _registryManager.CreateQuery(countSqlQuery);
+            while (countQuery.HasMoreResults)
+            {
+                IEnumerable<string> result = await countQuery.GetNextAsJsonAsync();
+                foreach (var item in result)
+                {
+                    Console.WriteLine($"Total # of devices in the hub: \n{item}");
```
bikamani, line 148: Is this foreach required? Just need to keep count.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,168 @@
...
+            _registryManager = rm ?? throw new ArgumentNullException(nameof(rm));
+            _blobContainerClient = sc ?? throw new ArgumentNullException(nameof(sc));
+
+            _deleteDeviceWithPrefix = deleteDeviceWithPrefix;
+        }
+
+        public async Task RunCleanUpAsync()
+        {
+            // Clean up devices
+            await PrintDeviceCountAsync();
```
bikamani, line 42: Does it make sense to print total devices count in hub or the number of devices with delete prefix (or to be deleted)?

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            _registryManager = rm ?? throw new ArgumentNullException(nameof(rm));
+            _blobContainerClient = sc ?? throw new ArgumentNullException(nameof(sc));
+
+            _deleteDeviceWithPrefix = deleteDeviceWithPrefix;
+        }
+
+        public async Task RunCleanUpAsync()
+        {
+            // Clean up devices
```
bikamani, line 46: I would reword this to clean up devices with _deleteDeviceWithPrefix.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
abhipsaMisra: the result is an IEnumerable, so we will need the for-each. For this specific query we know that result.Count = 1, but I didn't want to deviate from query-result patterns. That said, I don't have a strong preference for wanting to have it this way.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,173 @@
...
+            await PrintDeviceCountAsync();
+            await CleanupDevices();
+        }
+
+        private async Task CleanupDevices()
+        {
+            Console.WriteLine($"Storage container {_blobContainerClient.Name} created.");
```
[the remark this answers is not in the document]
abhipsaMisra: I am not sure what you mean here.
- Are you asking where the storage container is created? It is created within the storage account whose connection string you've passed in.
- Are you asking when it is created? It is created here: https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/abmisr/useJobForCleanup/iot-hub/Samples/service/CleanupDevicesSample/Program.cs#L41

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,173 @@
...
+            Console.WriteLine($"Job {exportAllDevicesJob.JobId} is {exportAllDevicesJob.Status}.");
+
+            if (exportAllDevicesJob.Status != JobStatus.Completed)
+            {
+                throw new Exception("Exporting devices failed, exiting.");
+            }
+
+            // Step 2: Download the exported devices list.
+            BlobClient blobClient = _blobContainerClient.GetBlobClient(ImportExportDevicesFileName);
+            BlobDownloadInfo download = await blobClient.DownloadAsync();
```
bikamani, line 89: What this line does wasn't self explanatory to me and had to look up.

Either adding a comment something like "Download the previously created blob" or renaming will add some more context.

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Parameter.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
     internal class Parameters
     {
         [Option(
-            's',
+            'f',
```
drwill-ms, line 17: 'f' for Flow?

drwill-ms: Btw, if we can't find a good single letter because of conflicts, we don't have to have one. I'd rather not have one than one that doesn't make obvious sense.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Samples
+{
+    /// <summary>
+    /// This class leverages the import and export IoT hub device identities feature to export all registered device identities,
+    /// filter out the devices that should not be deleted, and execute bulk deletion using ImportMode.Delete.
```
drwill-ms, line 17: Since ImportMode.Delete is a type, you can use the `<see>` tag.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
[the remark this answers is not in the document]
bikamani: Both answers help. But then it does not seem to be appropriate place for logging it.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
drwill-ms: Looks good.

drwill-ms (approved):

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
drwill-ms: Check with Robin Shahan.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
drwill-ms: I think you know it is going to be a single result, because you used an aggregate in the query string, so you can just grab the first item.

bikamani: Interesting, that was confusing at first. So the result count is always 1 but return type is IEnumerable<string>!!

Also it seems like we are printing count of all devices and not only the ones with _deleteDeviceWithPrefix, correct me if that's wrong.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
bikamani: If you want let's talk offline on this.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
bikamani: Ok so it not as part of this method PrintDeviceCountAsync, why do we then need this method? OR is is there to demonstrate how to use query API?

bikamani:
> _No description provided._

Since you may have added more changes than just clean up samples, it would be good to update description.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
abhipsaMisra: > Ok so it not as part of this method PrintDeviceCountAsync, why do we then need this method?

There are 2 counts in question here - the total no of devices on your hub and the no of devices being deleted. The reason we print both of those is for us to compare and verify that we are indeed deleting all "expected" devices that fit the criteria (device prefixes that are supplied as an input to this application). If we print only the count of devices being deleted, we wouldn't know if our input file/ deletion logic is incorrect unless we look at the hub instance and verify the device registry.

> OR is is there to demonstrate how to use query API?

This is not the intention, no.

bikamani: That would make sense if there is any logic of comparing these counts to validate. Shouldn't logging devices id which are being deleted also be sufficient since those are the ones with prefix? But either ways, I don't have a strong preference for the reasoning you mentioned so I'll leave it to you for decision.

If kept as it is, it would be ideal to rename it to PrintTotalDevicesCountAsync or add a comment for clarity with reasoning to why you think it needs to be printed. As you mentioned earlier .... to verify that we are indeed deleting all "expected" devices that fit the criteria (device prefixes that are supplied as an input to this application)

PR #192, opened 2021-04-14
Close all device client instances before disposing them

timtay-microsoft:
The best practice for safely disposing of a device client instance is to close it first, and then dispose it. These samples need to be updated to reflect that

### DIFF
--- a/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
         private static readonly TimeSpan s_sleepDuration = TimeSpan.FromSeconds(5);
 
-        private readonly object _initLock = new object();
+        private readonly SemaphoreSlim _initSemaphore = new SemaphoreSlim(1, 1);
         private readonly List<string> _deviceConnectionStrings;
         private readonly TransportType _transportType;
         private readonly ClientOptions _clientOptions = new ClientOptions { SdkAssignsMessageId = Shared.SdkAssignsMessageId.WhenUnset };
@@ -81,7 +81,8 @@ namespace Microsoft.Azure.Devices.Client.Samples
             if (ShouldClientBeInitialized(s_connectionStatus))
             {
                 // Allow a single thread to dispose and initialize the client instance.
-                lock (_initLock)
+                await _initSemaphore.WaitAsync();
+                try
                 {
                     if (ShouldClientBeInitialized(s_connectionStatus))
                     {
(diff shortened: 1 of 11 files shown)


iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -90,6 +90,7 @@ private async Task InitializeAndOpenClientAsync()
                         // If the device client instance has been previously initialized, then dispose it.
                         if (s_deviceClient != null)
                         {
+                            s_deviceClient.CloseAsync().GetAwaiter().GetResult(); // cannot await a task in a lock statement
```
drwill-ms, line 93: We could choose instead to use a semaphore. That would be better, I think.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -19,7 +19,7 @@ public class DeviceReconnectionSample
 
         private static readonly TimeSpan s_sleepDuration = TimeSpan.FromSeconds(5);
 
-        private readonly object _initLock = new object();
+        private readonly SemaphoreSlim _initSemaphore = new SemaphoreSlim(1, 1);
```
drwill-ms, line 22: This should be disposed when the DeviceReconnectionSample is done, right?

drwill-ms (approved):

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -90,6 +93,7 @@ private async Task InitializeAndOpenClientAsync()
                         // If the device client instance has been previously initialized, then dispose it.
                         if (s_deviceClient != null)
                         {
+                            await s_deviceClient.CloseAsync();
```
bikamani, line 96: Happen to notice, some places you have used a token and not others? 

await s_deviceClient.CloseAsync(cts.Token);

bikamani (approved):

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
timtay-microsoft: Some samples in this repo have a cancellation token for the sample, and others do not. When the sample already had a cancellation token, I passed it into the new call to deviceClient.CloseAsync accordingly

barustum (approved):

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
bikamani: Gotcha. For this one I remember seeing it though.

PR #194, opened 2021-04-27
Updating device SDK versions

drwill-ms:


iot-hub/Samples/device/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
     public class MessageReceiveSample
     {
         private static readonly TimeSpan s_sleepDuration = TimeSpan.FromSeconds(5);
-        private static readonly TimeSpan s_receiveTimeout = TimeSpan.FromSeconds(10);
```
drwill-ms, line 23: This was too fast to let someone who isn't 100% ready in the portal or IoT Explorer.

iot-hub/Samples/device/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -97,8 +97,8 @@ private void ProcessReceivedMessage(Message receivedMessage)
             {
                 formattedMessage.AppendLine($"\tProperty: key={prop.Key}, value={prop.Value}");
             }
-            // System properties can be accessed using their respective accessors.
-            formattedMessage.AppendLine($"\tMessageId: {receivedMessage.MessageId}");
+            // System properties can be accessed using their respective accessors, e.g. DeliveryCount.
+            formattedMessage.AppendLine($"\tDelivery count: {receivedMessage.DeliveryCount}");
```
drwill-ms, line 101: MessageId is not set from the portal, so I picked one that has a value.

iot-hub/Samples/device/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -71,6 +70,7 @@ private async Task ReceiveC2dMessagesPollingAndComplete(TimeSpan timeout)
 
                 await _deviceClient.CompleteAsync(receivedMessage);
                 Console.WriteLine($"{DateTime.Now}> Completed C2D message with Id={receivedMessage.MessageId}.");
+                break;
```
abhipsaMisra, line 73: this break isn't required, right? we want to keep polling until we time out, and not just on 1st message received.

abhipsaMisra (approved):

iot-hub/Samples/device/MessageReceiveSample/MessageReceiveSample.cs
drwill-ms: I found the sample flow confusing otherwise. I sent a message that was picked up via polling, and was hoping to move on to the next one. Instead I just get to wait.

Well, if you are waiting 10 seconds for the whole operation then it isn't too long, but as I mentioned in another comment, 10 seconds wasn't enough for me to get it done.

So I figured, giving you 1 minute to get a C2D message off, and then proceed to the next part of the sample was more interesting.

Thoughts?

abhipsaMisra: I can get on board with that. 
The idea initially was to show the common prod scenario, where applications would continuously poll for messages and process them as received. This not running on a separate thread but rather waiting for this polling to complete to continue to other scenarios kind-of deviates from that idea anyway.

I'd recommend updating the method name and associated comments to indicate that this method will process a single message:
Device waiting for C2D messages from the hub for {s_receiveTimeout}... -> Device waiting for a C2D message from the hub for {s_receiveTimeout}...
ReceiveC2dMessagesPollingAndComplete -> ReceiveC2dMessagePollingAndComplete

drwill-ms: What if we changed it to take user input to move to the next phase? That would mean no timeout. It says in Phase 1 until told to move to Phase 2, and then same in Phase 2 until time to quit?

abhipsaMisra: Yeah, that'd be good. I think that'll tie in well with the whole concept of the polling receive API.
Small note - the default case will need to have a timeout (customizable), just because we want this sample to run as a part of our nightly pipeline run.

jamdavi (approved):

iot-hub/Samples/device/MessageReceiveSample/MessageReceiveSample.cs
drwill-ms: Updated. Take a look! :)

iot-hub/Samples/device/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -19,75 +20,95 @@ namespace Microsoft.Azure.Devices.Client.Samples
...
+            {
+                await Task.Delay(-1, cts.Token);
+            }
+            catch (TaskCanceledException)
+            {
+                // Done running.
+            }
 
             // Now unsubscibe from receiving the callback.
-            await _deviceClient.SetReceiveMessageHandlerAsync(null, _deviceClient);
```
drwill-ms, line 49: I don't know why it was setting the device client as the context.

iot-hub/Samples/device/MessageReceiveSample/Parameters.cs
```diff
@@ -21,5 +21,12 @@ internal class Parameters
             Required = false,
             HelpText = "The transport to use to communicate with the IoT Hub. Possible values include Mqtt, Mqtt_WebSocket_Only, Mqtt_Tcp_Only, Amqp, Amqp_WebSocket_Only, Amqp_Tcp_only, and Http1.")]
         public TransportType TransportType { get; set; }
+
+        [Option(
+            'r',
```
drwill-ms, line 26: Now this sample supports the `-r` param, just like the reconnection sample. The default is null, which equates to run until explicitly canceled by the user using control+c.

iot-hub/Samples/device/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -19,75 +20,95 @@ namespace Microsoft.Azure.Devices.Client.Samples
...
             // Now wait to receive C2D messages through the callback.
             // Since you are subscribed to receive messages through the callback, any call to the polling ReceiveAsync() API will now return "null".
-            Console.WriteLine($"\n{DateTime.Now}> Device waiting for C2D messages from the hub for {s_receiveTimeout}...");
+            Console.WriteLine($"\n{DateTime.Now}> Device waiting for C2D messages from the hub...");
             Console.WriteLine($"{DateTime.Now}> Use the Azure Portal IoT Hub blade or Azure IoT Explorer to send a message to this device.");
-            await Task.Delay(s_receiveTimeout);
+
+            try
+            {
+                await Task.Delay(-1, cts.Token);
```
abhipsaMisra, line 62: by the time we reach this point cts.Token would already have been cancelled, right? this would require a new cancellation token to be created.

abhipsaMisra: ah, N is to move to the next phase, I see

iot-hub/Samples/device/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -19,75 +20,95 @@ namespace Microsoft.Azure.Devices.Client.Samples
...
-            sw.Start();
+            Console.WriteLine($"{DateTime.Now}> Trying to receive C2D messages by polling using the ReceiveAsync() method. Press 'n' to move to the next phase.");
 
-            Console.WriteLine($"{DateTime.Now}> Receiving C2D messages on the polling ReceiveAsync().");
-            while (sw.Elapsed < timeout)
+            while (!ct.IsCancellationRequested)
             {
-                using Message receivedMessage = await _deviceClient.ReceiveAsync(timeout);
+                if (Console.KeyAvailable
+                    && ConsoleKey.N == Console.ReadKey().Key)
```
abhipsaMisra, line 80: is this case-sensitive?

abhipsaMisra (approved):

iot-hub/Samples/device/MessageReceiveSample/MessageReceiveSample.cs
abhipsaMisra: how to manage this "move" in the pipeline?

abhipsaMisra: The pipeline build script needs to be updated: https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/build.ps1#L115

drwill-ms: It is not. :)

iot-hub/Samples/device/MessageReceiveSample/MessageReceiveSample.cs
drwill-ms: Why would it have been signaled? That only happens if we run out of time or the user presses control-c.

In this case, we're on the second part of the sample, using subscribe, but we don't have any code we're waiting on (because it is a callback).

PR #200, opened 2021-04-30
refactor(sample): Simplifying the import export job sample with msi.

vinagesh:
## Purpose

* ...

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


iot-hub/Samples/service/ImportExportDevicesWithManagedIdentitySample/Parameters.cs
```diff
@@ -28,16 +28,10 @@ internal class Parameters
             HelpText = "The Uri of storage container for import and export jobs.")]
         public string BlobContainerUri { get; set; }
 
-        [Option(
```
drwill-ms, line 31: We're removing the option to run this sample without any kind of identity?

vinagesh: This was something I added earlier. This is not something the API requires. It just makes the sample harder to understand. The sample explicitly says this is for managed identites only. So the StorageAuthenticationType is set to IdentityBased in JobProperties.

vinagesh: The API does not ask if you want to do system-defined or user-defined. That is dictated by the fact that Identity property is present or not on the request body. I have documented that in detail.

drwill-ms (approved):

PR #209, opened 2021-06-08
doc(device-sample): Added readme and enhanced setup script for x509 cert chain sample

vinagesh:
## Purpose

* ...

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/iot-hub/Samples/device/X509DeviceCertWithChainSample/Program.cs
+++ b/iot-hub/Samples/device/X509DeviceCertWithChainSample/Program.cs
@@ -47,7 +47,6 @@ namespace X509DeviceCertWithChainSample
 
             var sample = new X509DeviceCertWithChainSample(deviceClient);
             await sample.RunSampleAsync();
-            await deviceClient.CloseAsync();
 
             return 0;
         }
--- a/iot-hub/Samples/device/X509DeviceCertWithChainSample/readme.md
+++ b/iot-hub/Samples/device/X509DeviceCertWithChainSample/readme.md
@@ -1,0 +1,44 @@
+# x509 device certificate with chain sample
+
+## How to run the sample
+
+### Step 1
+
+Create a device in your IoT hub using the x.509 CA signed option.
+
+### Step 2
+
+Create a secure string password to use in creation of self-signed certificates.
+
+```powershell
+    $password= ConvertTo-SecureString <your password> -AsPlainText -Force
+```
+
+### Step 3
+
+Run the [AuthSetup.ps1](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/X509DeviceCertWithChainSample/AuthSetup.ps1) to setup the IoT hub and device with necessary certificates.
+
+```powershell
+    .\AuthSetup.ps1 `
+    -certFolderPath <Path to root cert> `
+    -rootCertPassword $password `
+    -iotHubResourceGroup <IoT hub resource group> `
+    -iotHubName <IoT hub name> `
+    -deviceId <device Id> 
+```
+
+### Step 4
+
+Run the sample with the following command line arguments.
+
+```
+    "commandLineArgs": "-h <hostname of IoT hub> 
+                        -d <device Id> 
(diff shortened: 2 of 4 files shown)


iot-hub/Samples/device/X509DeviceCertWithChainSample/readme.md
```diff
@@ -0,0 +1,21 @@
+# x509 device certificate with chain sample
+
+## How to run the sample
+
+1) Create IotHub and device to use in sample.
```
barustum, line 5: shouldn't creating a device be after you run the AuthSetup in order to grab the thumbprint that is generated by the script and use it when creating a device?

iot-hub/Samples/device/X509DeviceCertWithChainSample/readme.md
```diff
@@ -0,0 +1,21 @@
...
+-iotHubResourceGroup <IoT hub resource group> `
+-iotHubName <IoT hub name> `
+-deviceId <device name> 
+```
+
+## Important things to note
+
+1) Use the same IoT hub created in step 1 as input to step 2 and step 3.
+2) Use the same device created in step 1 as input to step 2 and step 3.
+3) Use the same password created in in step 2 as input to step 3.
```
barustum, line 21: When I ran the script and provided my own password, it complained that:
AuthSetup.ps1: Cannot process argument transformation on parameter 'rootCertPassword'. Cannot convert the "password" value of type "System.String" to type "System.Security.SecureString".

I had to create a plain password then call Convert-ToSecureString then pass that to the script.

vinagesh: Yes that is what is expected as it is a secure string input. But it would be good to call it out with an example in the readme.

iot-hub/Samples/device/X509DeviceCertWithChainSample/readme.md
```diff
@@ -0,0 +1,21 @@
...
+-certFolderPath <Path to root cert> `
+-rootCertPassword <root certificate password> `
+-iotHubResourceGroup <IoT hub resource group> `
+-iotHubName <IoT hub name> `
+-deviceId <device name> 
+```
+
+## Important things to note
+
+1) Use the same IoT hub created in step 1 as input to step 2 and step 3.
```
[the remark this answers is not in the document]
vinagesh: It's there below the powershell snippet. But let me add some details on inputs to step 3

iot-hub/Samples/device/X509DeviceCertWithChainSample/readme.md
vinagesh: Not needed if you use the x.509 CA signed option. It is simpler to use that option. But both options will work fine. Though you are correct of the order of things if you use the self signed option.

iot-hub/Samples/device/X509DeviceCertWithChainSample/readme.md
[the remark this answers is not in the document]
vinagesh: I updated the instructions some more to add clarity.

iot-hub/Samples/device/X509DeviceCertWithChainSample/Program.cs
```diff
@@ -47,7 +47,6 @@ public static async Task<int> Main(string[] args)
 
             var sample = new X509DeviceCertWithChainSample(deviceClient);
             await sample.RunSampleAsync();
-            await deviceClient.CloseAsync();
```
bikamani, line 50: why was this removed?

vinagesh: It is already there in the sample run, not program.cs. I think someone added it here again and the sample was failing as we were calling Close twice.

barustum (approved):

PR #222, opened 2021-07-20
Minor sample improvements for consistency

vinagesh:
Minor sample improvements for consistency

### DIFF
--- a/iot-hub/Samples/service/RegistryManagerSample/RegistryManagerSample.cs
+++ b/iot-hub/Samples/service/RegistryManagerSample/RegistryManagerSample.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Devices.Samples
         public async Task RunSampleAsync()
         {
             using RegistryManager registryManager = RegistryManager
-                .CreateFromConnectionString(_parameters.ConnectionString);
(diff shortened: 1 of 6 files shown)


abhipsaMisra (approved):

iot-hub/Samples/service/RegistryManagerSample/RegistryManagerSample.cs
```diff
@@ -189,13 +189,13 @@ private async Task UpdateDesiredPropertiesAsync(RegistryManager registryManager,
...
             var patch =
-@"{
-  ""properties"": {
-    ""desired"": {
-      ""customKey"": ""customValue""
-    }
-  }
-}";
+            @"{
+              ""properties"": {
```
drwill-ms, line 193: This will insert a bunch of additional spaces in the payload. Do we want to do that?

drwill-ms (approved):

iot-hub/Samples/service/RegistryManagerSample/RegistryManagerSample.cs
vinagesh: Probably not but the sample works even with the spaces. I'll revert this.

PR #227, opened 2021-08-20
 Do not delete e2e test resources on PR builds

vinagesh:
This PR mainly moves the configuration of which devices to delete to the pipeline variable rather than a constant value. Once this is checked in, the scheduled run will run as normal, running all tests and cleaning up both e2e and sample created devices.

Apart from that pipeline, I"ll add a new pipeline that will only delete the sample created devices along with running the samples. This pipeline will be enabled on PR builds.

After this is complete, I will make merge these changes to preview branch and create a PR pipeline for the preview branch as well.

### DIFF
--- a/iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
+++ b/iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
@@ -4,6 +4,8 @@
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
 using Azure.Storage.Sas;
+using Microsoft.Azure.Devices.Common.Exceptions;
+using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -20,6 +22,8 @@ namespace Microsoft.Azure.Devices.Samples
     public class CleanupDevicesSample
     {
         private const string ImportExportDevicesFileName = "devices.txt";
+        private const int MaxIterationWait = 60;
+        private static readonly TimeSpan WaitDuration = TimeSpan.FromSeconds(10);
 
         private static readonly TimeSpan s_waitDuration = TimeSpan.FromSeconds(5);
         private static readonly IReadOnlyList<JobStatus> s_completedJobs = new[]
@@ -37,7 +41,7 @@ namespace Microsoft.Azure.Devices.Samples
         {
             _registryManager = rm ?? throw new ArgumentNullException(nameof(rm));
             _blobContainerClient = sc ?? throw new ArgumentNullException(nameof(sc));
-
+            Console.WriteLine($"Delete devices with prefixes: {JsonConvert.SerializeObject(deleteDevicesWithPrefix)}");
             _deleteDevicesWithPrefix = deleteDevicesWithPrefix;
         }
 
@@ -65,8 +69,29 @@ namespace Microsoft.Azure.Devices.Samples
                     excludeKeysInExport: true,
                     storageAuthenticationType: StorageAuthenticationType.KeyBased);
 
-            JobProperties exportAllDevicesJob = await _registryManager.ExportDevicesAsync(exportAllDevicesProperties);
+            JobProperties exportAllDevicesJob = null;
+            
+            int tryCount = 0;
+            while (true)
+            {
+                try
+                {
+                    exportAllDevicesJob = await _registryManager.ExportDevicesAsync(exportAllDevicesProperties);
+                    break;
+                }
+                // Wait for pending jobs to finish.
+                catch (JobQuotaExceededException) when (++tryCount < MaxIterationWait)
+                {
+                    Console.WriteLine($"JobQuotaExceededException... waiting.");
+                    await Task.Delay(WaitDuration);
+                }
+            }
 
+            if (exportAllDevicesJob == null)
+            {
+                throw new Exception("Export devices job failed.");
+            }
+            
             // Wait until the export job is finished.
             while (true)
             {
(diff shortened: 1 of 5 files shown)


iot-hub/Samples/service/RegistryManagerSample/Parameters.cs
```diff
@@ -29,7 +29,7 @@ internal class Parameters
             'd',
             "DevicePrefix",
             Required = false,
-            Default = "RegistryManagerSample-",
```
vinagesh, line 32: Changing to match prefix. This was not in the delete list earlier but the sample deletes the device so this is just to guard against leaked resources. Currently the hub does not have any devices with this prefix so we don't need to worry about deleting the ones with old prefix.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -65,8 +69,29 @@ private async Task CleanupDevices()
                     excludeKeysInExport: true,
                     storageAuthenticationType: StorageAuthenticationType.KeyBased);
 
-            JobProperties exportAllDevicesJob = await _registryManager.ExportDevicesAsync(exportAllDevicesProperties);
+            JobProperties exportAllDevicesJob = null;
```
vinagesh, line 72: Adding retries on jobs to protect against any currently running job. We could change the total wait time if necessary after observing for a while.

vsts/vsts.yaml
```diff
@@ -53,6 +53,10 @@ jobs:
           IOTHUB_PFX_X509_THUMBPRINT: $(IOTHUB-PFX-X509-THUMBPRINT)
           DPS_IDSCOPE: $(DPS-IDSCOPE)
           PROVISIONING_CONNECTION_STRING: $(PROVISIONING-CONNECTION-STRING)
+          # PATH-TO-DEVICE-PREFIX-FOR-DELETION-FILE is a variable configured on the pipeline. The possible values are: csharp_devices_list.csv, csharp_samples_devices_list.csv.
```
vinagesh, line 56: This variable is no longer a constant and no longer extracted from keyvault.

timtay-microsoft (approved):

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -37,7 +41,7 @@ public CleanupDevicesSample(RegistryManager rm, BlobContainerClient sc, List<str
         {
             _registryManager = rm ?? throw new ArgumentNullException(nameof(rm));
             _blobContainerClient = sc ?? throw new ArgumentNullException(nameof(sc));
-
+            Console.WriteLine($"Delete devices with prefixes: {JsonConvert.SerializeObject(deleteDevicesWithPrefix)}");
```
drwill-ms, line 44: Why call `JsonConvert.SerializeObject()` on a list of strings?

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -65,8 +69,29 @@ private async Task CleanupDevices()
...
 
-            JobProperties exportAllDevicesJob = await _registryManager.ExportDevicesAsync(exportAllDevicesProperties);
+            JobProperties exportAllDevicesJob = null;
+            
+            int tryCount = 0;
+            while (true)
+            {
+                try
+                {
+                    exportAllDevicesJob = await _registryManager.ExportDevicesAsync(exportAllDevicesProperties);
```
drwill-ms, line 79:
```suggestion
                    exportAllDevicesJob = await _registryManager.ExportDevicesAsync(exportAllDevicesProperties).ConfigureAwait(false);

```

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -65,8 +69,29 @@ private async Task CleanupDevices()
...
+                }
+                // Wait for pending jobs to finish.
+                catch (JobQuotaExceededException) when (++tryCount < MaxIterationWait)
+                {
+                    Console.WriteLine($"JobQuotaExceededException... waiting.");
+                    await Task.Delay(WaitDuration).ConfigureAwait(false);
+                }
+            }
 
+            if(exportAllDevicesJob == null)
```
drwill-ms, line 90:
```suggestion
            if (exportAllDevicesJob == null)

```

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
```diff
@@ -126,7 +151,29 @@ private async Task CleanupDevices()
...
-                JobProperties importDevicesToBeDeletedJob = await _registryManager.ImportDevicesAsync(importDevicesToBeDeletedProperties);
+
+                JobProperties importDevicesToBeDeletedJob = null;
+
+                tryCount = 0;
+                while (true)
+                {
+                    try
+                    {
+                        importDevicesToBeDeletedJob = await _registryManager.ImportDevicesAsync(importDevicesToBeDeletedProperties);
```
drwill-ms, line 162:
```suggestion
                        importDevicesToBeDeletedJob = await _registryManager
                            .ImportDevicesAsync(importDevicesToBeDeletedProperties)
                            .ConfigureAwait(false);

```

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
vinagesh: I just want to print all the values in the list. I could use string.Join or linq but just chose one way to print. If there is an argument for why one way is better that the other, I can change it.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
vinagesh: From what I know, we don't use configureawait in samples.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
vinagesh: From what I know, we don't use ConfigureAwait in samples.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
drwill-ms: Ah, I see. Yeah, I was a bit confused but now I get it. I think string.join would be more straight forward, but I think this is acceptable as well.

iot-hub/Samples/service/CleanupDevicesSample/CleanupDevicesSample.cs
drwill-ms: You are right. We do have `ConfigureAwait()` elsewhere in this file, though, so we should clean that up then.

vinagesh: Will clean up!

drwill-ms (approved):

PR #228, opened 2021-08-24
add sleep command to fix Add-AzIotHubRoutingEndpoint 

robinsh:
Add the ability to put in a sleep comman to account for the timing issue with powershell commands used.


robinsh (commented):
Update powershell command.

drwill-ms (approved):

robinsh:
Do I have to #sign-off again?

PR #235, opened 2021-09-03
Add twin operations to ReconnectionSample

abhipsaMisra:
This PR brings two updates:
1. Consolidate the "common" retry logic into a retry helper.
2. Add twin desired property update + reported property update flow.

This PR does not address the [client reconnection](https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-device-twins#device-reconnection-flow) concerns with the twin update flow yet. I'll add it in a separate PR once we have the resolution on what the expectation is supposed to be.

### DIFF
--- a/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
@@ -112,12 +112,9 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 try
                 {
                     // Force connection now.
-                    // OpenAsync() is an idempotent call, it has the same effect if called once or multiple times on the same client.
-                    await s_deviceClient.OpenAsync(cancellationToken);
-                    _logger.LogDebug("Opened the client instance.");
-
+                    // Client operations have implicit open enabled.
                     await SubscribeToTwinUpdateNotificationsAsync(cancellationToken);
-                    _logger.LogDebug("Subscribed to desired property update notifications");
+                    _logger.LogDebug("Opened the client connection and subscribed to desired property update notifications");
                 }
                 catch (UnauthorizedException)
                 {
@@ -222,7 +219,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
             {
                 var reportedProperties = new TwinCollection();
 
-                _logger.LogInformation($"Twin properties update requested: \n{twinUpdateRequest.ToJson()}");
+                _logger.LogInformation($"Twin property update requested: \n{twinUpdateRequest.ToJson()}");
 
                 // For the purpose of this sample, we'll blindly accept all twin property write requests.
                 foreach (KeyValuePair<string, object> desiredProperty in twinUpdateRequest)
--- a/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
@@ -68,8 +68,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
             try
             {
-                await InitializeAndOpenClientAsync();
-                await SubscribeToTwinUpdateNotificationsAsync(cts.Token);
+                await InitializeAndSetupClientAsync(cts.Token);
                 await Task.WhenAll(SendMessagesAsync(cts.Token), ReceiveMessagesAsync(cts.Token));
             }
             catch (Exception ex)
@@ -80,7 +79,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
             _initSemaphore.Dispose();
         }
 
-        private async Task InitializeAndOpenClientAsync()
+        private async Task InitializeAndSetupClientAsync(CancellationToken cancellationToken = default)
         {
             if (ShouldClientBeInitialized(s_connectionStatus))
             {
@@ -103,7 +102,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
                     s_deviceClient = DeviceClient.CreateFromConnectionString(_deviceConnectionStrings.First(), _transportType, _clientOptions);
                     s_deviceClient.SetConnectionStatusChangesHandler(ConnectionStatusChangeHandler);
-                    _logger.LogDebug($"Initialized the client instance.");
+                    _logger.LogDebug("Initialized the client instance.");
                 }
                 finally
                 {
@@ -114,8 +113,11 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 {
                     // Force connection now.
                     // OpenAsync() is an idempotent call, it has the same effect if called once or multiple times on the same client.
-                    await s_deviceClient.OpenAsync();
-                    _logger.LogDebug($"Opened the client instance.");
+                    await s_deviceClient.OpenAsync(cancellationToken);
+                    _logger.LogDebug("Opened the client instance.");
+
+                    await SubscribeToTwinUpdateNotificationsAsync(cancellationToken);
+                    _logger.LogDebug("Subscribed to desired property update notifications");
                 }
                 catch (UnauthorizedException)
                 {
@@ -158,7 +160,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
                             if (_deviceConnectionStrings.Any())
                             {
                                 _logger.LogWarning($"The current connection string is invalid. Trying another.");
-                                await InitializeAndOpenClientAsync();
+                                await InitializeAndSetupClientAsync();
                                 break;
                             }
 
@@ -174,14 +176,14 @@ namespace Microsoft.Azure.Devices.Client.Samples
                             _logger.LogWarning("### The DeviceClient has been disconnected because the retry policy expired." +
                                 "\nIf you want to perform more operations on the device client, you should dispose (DisposeAsync()) and then open (OpenAsync()) the client.");
 
-                            await InitializeAndOpenClientAsync();
+                            await InitializeAndSetupClientAsync();
                             break;
 
                         case ConnectionStatusChangeReason.Communication_Error:
                             _logger.LogWarning("### The DeviceClient has been disconnected due to a non-retry-able exception. Inspect the exception for details." +
                                 "\nIf you want to perform more operations on the device client, you should dispose (DisposeAsync()) and then open (OpenAsync()) the client.");
 
-                            await InitializeAndOpenClientAsync();
+                            await InitializeAndSetupClientAsync();
                             break;
 
                         default:
@@ -220,13 +222,12 @@ namespace Microsoft.Azure.Devices.Client.Samples
             {
                 var reportedProperties = new TwinCollection();
 
-                Console.WriteLine("\nTwin properties update requested:");
-                Console.WriteLine($"\t{twinUpdateRequest.ToJson()}");
+                _logger.LogInformation($"Twin properties update requested: \n{twinUpdateRequest.ToJson()}");
 
                 // For the purpose of this sample, we'll blindly accept all twin property write requests.
                 foreach (KeyValuePair<string, object> desiredProperty in twinUpdateRequest)
                 {
-                    Console.WriteLine($"Setting {desiredProperty.Key} to {desiredProperty.Value}.");
+                    _logger.LogInformation($"Setting property {desiredProperty.Key} to {desiredProperty.Value}.");
                     reportedProperties[desiredProperty.Key] = desiredProperty.Value;
                 }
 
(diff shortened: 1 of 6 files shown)


iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -197,6 +197,47 @@ private async void ConnectionStatusChangeHandler(ConnectionStatus status, Connec
             }
         }
 
+        private async Task SubscribeToTwinUpdateNotificationsAsync(CancellationToken cancellationToken)
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                if (IsDeviceConnected)
+                {
+                    await s_deviceClient.SetDesiredPropertyUpdateCallbackAsync(HandleTwinUpdateNotificationsAsync, cancellationToken);
```
timtay-microsoft, line 206: Why does this operation need to be in a while loop? Wouldn't one check of the cancellation token and one attempt at setting the desired property callback be enough?

timtay-microsoft (approved):

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: Ideally, yes, but we don't want to set the callback unless the client is connected.

drwill-ms: Wouldn't that have been solved by keeping the explicit call to OpenAsync?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -197,6 +197,47 @@ private async void ConnectionStatusChangeHandler(ConnectionStatus status, Connec
...
+                foreach (KeyValuePair<string, object> desiredProperty in twinUpdateRequest)
+                {
+                    _logger.LogInformation($"Setting property {desiredProperty.Key} to {desiredProperty.Value}.");
+                    reportedProperties[desiredProperty.Key] = desiredProperty.Value;
+                }
+
+                await RetryOperationHelper.RetryTransientExceptionsAsync(
+                        async () =>
+                        {
+                            await s_deviceClient.UpdateReportedPropertiesAsync(reportedProperties, cancellationToken);
```
drwill-ms, line 234: Requires a comment that a device probably wouldn't blindly accept property updates and should do some checking, but for the sake of simplicity we do not do validation in this sample.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -345,6 +342,23 @@ private async Task ReceiveMessageAndCompleteAsync()
             }
         }
 
+        private (Message, string) PrepareMessage(int messageId)
```
drwill-ms, line 345: This will make disposing `Message` difficult because of the tuple. Thoughts?

iot-hub/Samples/device/DeviceReconnectionSample/RetryOperationHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+
+using Microsoft.Azure.Devices.Client.Exceptions;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Client.Samples
+{
+    internal class RetryOperationHelper
```
drwill-ms, line 12: Could use some meaningful doc comments.

iot-hub/Samples/device/DeviceReconnectionSample/RetryOperationHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (retryPolicy == null)
+            {
+                retryPolicy = s_exponentialBackoffRetryStrategy;
+            }
+
+            int counter = 0;
+            bool shouldRetry;
+            do
+            {
+                Exception lastException = new IotHubCommunicationException("Client is currently reconnecting internally, attempt the operation after some time.");
```
drwill-ms, line 36:
```suggestion
                Exception lastException = new IotHubCommunicationException("Client is currently reconnecting internally; attempt the operation after some time.");

```

iot-hub/Samples/device/DeviceReconnectionSample/RetryOperationHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+                try
+                {
+                    if (isClientConnected())
+                    {
+                        await asyncOperation().ConfigureAwait(false);
```
drwill-ms, line 41:
```suggestion
                        await asyncOperation();

```

iot-hub/Samples/device/DeviceReconnectionSample/RetryOperationHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+
+                    lastException = ex;
+                }
+
+                shouldRetry = retryPolicy.ShouldRetry(++counter, lastException, out TimeSpan retryInterval);
+
+                if (shouldRetry)
+                {
+                    logger.LogInformation($"Will retry operation in {retryInterval}, attempt {counter}.");
+                    await Task.Delay(retryInterval).ConfigureAwait(false);
```
drwill-ms, line 73:
```suggestion
                    await Task.Delay(retryInterval);

```

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: Not really, disconnections can happen at any point, right

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: Do you think moving the comment in L224 to the reported API call would make this clearer?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: Actually we don't seem to be using the payload in our sample. I'll update this.
We were disposing the message once the "send" operation was complete, after this change we can use `using`

iot-hub/Samples/device/DeviceReconnectionSample/RetryOperationHelper.cs
abhipsaMisra: sure

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: Yes, best make it as contextually relevant to where the check would have happened.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -112,9 +112,9 @@ private async Task InitializeAndOpenClientAsync()
                 try
                 {
                     // Force connection now.
-                    // OpenAsync() is an idempotent call, it has the same effect if called once or multiple times on the same client.
-                    await s_deviceClient.OpenAsync();
-                    _logger.LogDebug($"Opened the client instance.");
+                    // Client operations have implicit open enabled.
+                    await SubscribeToTwinUpdateNotificationsAsync(cancellationToken);
```
abhipsaMisra, line 116: add back the open call and add a note calling out subscriptions need to be added again

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -197,6 +197,47 @@ private async void ConnectionStatusChangeHandler(ConnectionStatus status, Connec
...
         }
 
+        private async Task SubscribeToTwinUpdateNotificationsAsync(CancellationToken cancellationToken)
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                if (IsDeviceConnected)
+                {
+                    await s_deviceClient.SetDesiredPropertyUpdateCallbackAsync(HandleTwinUpdateNotificationsAsync, cancellationToken);
+                    break;
```
abhipsaMisra, line 207: ensure that the task is cancelled if the client ends up being disconnected

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -56,30 +57,31 @@ public DeviceReconnectionSample(List<string> deviceConnectionStrings, TransportT
 
         public async Task RunSampleAsync(TimeSpan sampleRunningTime)
         {
-            using var cts = new CancellationTokenSource(sampleRunningTime);
+            s_cancellationTokenSource = new CancellationTokenSource(sampleRunningTime);
```
abhipsaMisra, line 60: creating a static variable so that we can access the cancellation token from within the status change callback

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -14,22 +15,22 @@ namespace Microsoft.Azure.Devices.Client.Samples
...
         private readonly TransportType _transportType;
-        private readonly ClientOptions _clientOptions = new ClientOptions { SdkAssignsMessageId = Shared.SdkAssignsMessageId.WhenUnset };
+        private readonly ClientOptions _clientOptions = new ClientOptions { SdkAssignsMessageId = SdkAssignsMessageId.WhenUnset };
 
         private readonly ILogger _logger;
 
         // Mark these fields as volatile so that their latest values are referenced.
         private static volatile DeviceClient s_deviceClient;
         private static volatile ConnectionStatus s_connectionStatus = ConnectionStatus.Disconnected;
+        private static volatile CancellationTokenSource s_cancellationTokenSource;
```
drwill-ms, line 33: When does this get disposed?

drwill-ms: I see, bottom of the `RunSampleAsync method.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: What is the scope of things that will stop happening for this CTS? Perhaps we could get a better, more descriptive name?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -56,30 +57,32 @@ public DeviceReconnectionSample(List<string> deviceConnectionStrings, TransportT
...
             {
-                await InitializeAndOpenClientAsync();
-                await Task.WhenAll(SendMessagesAsync(cts.Token), ReceiveMessagesAsync(cts.Token));
+                await InitializeAndSetupClientAsync(s_cancellationTokenSource.Token);
+                await Task.WhenAll(SendMessagesAsync(s_cancellationTokenSource.Token), ReceiveMessagesAsync(s_cancellationTokenSource.Token));
             }
             catch (Exception ex)
             {
                 _logger.LogError($"Unrecoverable exception caught, user action is required, so exiting...: \n{ex}");
+                s_cancellationTokenSource.Cancel();
```
drwill-ms, line 78: Probably need to wait for the above tasks to complete before disposing and moving on, no?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -98,28 +101,47 @@ private async Task InitializeAndOpenClientAsync()
                             s_deviceClient.Dispose();
                             s_deviceClient = null;
                         }
-                    }
 
-                    s_deviceClient = DeviceClient.CreateFromConnectionString(_deviceConnectionStrings.First(), _transportType, _clientOptions);
-                    s_deviceClient.SetConnectionStatusChangesHandler(ConnectionStatusChangeHandler);
-                    _logger.LogDebug($"Initialized the client instance.");
+                        s_deviceClient = DeviceClient.CreateFromConnectionString(_deviceConnectionStrings.First(), _transportType, _clientOptions);
+                        s_deviceClient.SetConnectionStatusChangesHandler(ConnectionStatusChangeHandler);
```
drwill-ms, line 106: Shouldn't we pass the cancellation token into methods that could be canceled?

drwill-ms: Never mind.

drwill-ms (approved):

PR #249, opened 2021-12-17
Evolve .NET testing targets (add net6.0)

drwill-ms:
netcoreapp2.1 is out of support, and we should plan for its removal but not until we've retargeted our samples. net6.0 has released and is LTS. Let's remove the former and add the latter for our samples testing.


vsts/vsts.yaml
```diff
@@ -14,15 +14,15 @@ jobs:
     timeoutInMinutes: 60
     steps:
       - task: UseDotNet@2
-        displayName: 'Use .NET Core 5.0 SDK'
+        displayName: 'Use .NET 6.0 SDK'
```
drwill-ms, line 17: v6 and 5 are no longer named "core".

jamdavi (approved):

PR #252, opened 2022-01-12
Remove deprecated logging property - Fix cancellation token usage in AMQP

azabbasi:


iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -92,7 +91,7 @@ private async Task InitializeAndSetupClientAsync(CancellationToken cancellationT
             if (ShouldClientBeInitialized(s_connectionStatus))
             {
                 // Allow a single thread to dispose and initialize the client instance.
-                await _initSemaphore.WaitAsync();
+                await _initSemaphore.WaitAsync(cancellationToken);
```
drwill-ms, line 94: May need to handle an exception that is thrown if this is cancelled.

drwill-ms: Same with the calls below.

iot-hub/Samples/device/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -77,7 +77,7 @@ private static ILogger InitializeConsoleDebugLogger()
             {
                 builder
                 .AddFilter(level => level >= LogLevel.Debug)
-                .AddConsole(options =>
+                .AddSystemdConsole(options =>
```
drwill-ms, line 80: Can you say more about this change?

azabbasi: https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/obsolete-consoleloggeroptions-properties

ConsoleLoggerOptions properties have been deprecated, it's recommended to use ConsoleFormatterOptions properties instead.

azabbasi: This was the only warning that started popping up on the Device samples that I had to fix. back to 0.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
azabbasi: I think it's best if catch that exception at the program entry and log it. let me what you think.

drwill-ms (approved):

PR #255, opened 2022-02-11
Update DeviceReconnectionSample

brycewang-microsoft:
## Purpose

* Add features:
1. Get twin values when the device client is initialzed.
2. Validate if the desired version is outdated and process properties further.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
@@ -273,8 +273,11 @@ namespace Microsoft.Azure.Devices.Client.Samples
                     TwinCollection twinCollection = twin.Properties.Desired;
                     long currentDesiredVersion = twinCollection.Version;
 
+                    // Validation of outdated version happens here.
                     if (currentDesiredVersion > s_latestDesiredVersion)
                     {
+                        _logger.LogDebug($"The desired property version was changed from {s_latestDesiredVersion} to {currentDesiredVersion}.");
+
                         await RetryOperationHelper.RetryTransientExceptionsAsync(
                             operationName: "GetTwinUpdateActively",
                             asyncOperation: async () => await HandlePropertiesAsync(twinCollection, cancellationToken),
--- a/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
@@ -35,6 +35,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
         private static volatile ConnectionStatus s_connectionStatus = ConnectionStatus.Disconnected;
 
         private static CancellationTokenSource s_cancellationTokenSource;
+
         // A safe initial value for caching the twin desired properties version is 1, so the client
         // will process all previous property change requests and initialize the device application
         // after which this version will be updated to that, so we have a high water mark of which version number
--- a/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
@@ -124,13 +124,9 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 // Force connection now.
                 // We have set the "shouldExecuteOperation" function to always try to open the connection.
                 // OpenAsync() is an idempotent call, it has the same effect if called once or multiple times on the same client.
-                await RetryOperationHelper<Task>.RetryTransientExceptionsAsync(
+                await RetryOperationHelper.RetryTransientExceptionsAsync(
                     operationName: "OpenConnection",
-                    asyncOperation: async () => 
-                    {
-                        await s_deviceClient.OpenAsync(cancellationToken);
-                        return Task.FromResult(true);
-                    },
+                    asyncOperation: async () => await s_deviceClient.OpenAsync(cancellationToken),
                     shouldExecuteOperation: () => true,
                     logger: _logger,
                     exceptionsToBeIgnored: _exceptionsToBeIgnored,
@@ -139,7 +135,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
                 // The client retrieves twin values when it is initialized.
                 Twin twin = null;
-                await RetryOperationHelper<Twin>.RetryTransientExceptionsAsync(
+                await RetryOperationHelper.RetryTransientExceptionsAsync(
                     operationName: "GetTwin",
                     asyncOperation: async () => 
                     {
@@ -153,8 +149,6 @@ namespace Microsoft.Azure.Devices.Client.Samples
                             _logger.LogDebug($"The desired property version was changed from {s_localDesiredPropertyVersion} to {serverDesiredPropertyVersion}.");
                             await HandleTwinUpdateNotificationsAsync(twinCollection, cancellationToken);
                         }
-
-                        return twin;
                     },
                     shouldExecuteOperation: () => IsDeviceConnected,
                     logger: _logger,
@@ -164,13 +158,9 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 _logger.LogDebug("The client has retrieved the twin state after establishing connection to IoT hub. ");
 
                 // You will need to subscribe to the client callbacks any time the client is initialized.
-                await RetryOperationHelper<Task>.RetryTransientExceptionsAsync(
+                await RetryOperationHelper.RetryTransientExceptionsAsync(
                     operationName: "SubscribeTwinUpdates",
-                    asyncOperation: async () => 
-                    {
-                        await s_deviceClient.SetDesiredPropertyUpdateCallbackAsync(HandleTwinUpdateNotificationsAsync, cancellationToken);
-                        return Task.FromResult(true);
-                    },
+                    asyncOperation: async () => await s_deviceClient.SetDesiredPropertyUpdateCallbackAsync(HandleTwinUpdateNotificationsAsync, cancellationToken),
                     shouldExecuteOperation: () => IsDeviceConnected,
                     logger: _logger,
                     exceptionsToBeIgnored: _exceptionsToBeIgnored,
@@ -272,16 +262,12 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 }
 
                 s_localDesiredPropertyVersion = twinUpdateRequest.Version;
-                _logger.LogDebug($"The desried property version on local is currently {s_localDesiredPropertyVersion}.");
+                _logger.LogDebug($"The desired property version on local is currently {s_localDesiredPropertyVersion}.");
 
                 // For the purpose of this sample, we'll blindly accept all twin property write requests.
-                await RetryOperationHelper<Task>.RetryTransientExceptionsAsync(
+                await RetryOperationHelper.RetryTransientExceptionsAsync(
                     operationName: "UpdateReportedProperties",
-                    asyncOperation: async () =>
-                    {
-                        await s_deviceClient.UpdateReportedPropertiesAsync(reportedProperties, cancellationToken);
-                        return Task.FromResult(true);
-                    },
+                    asyncOperation: async () => await s_deviceClient.UpdateReportedPropertiesAsync(reportedProperties, cancellationToken),
                     shouldExecuteOperation: () => IsDeviceConnected,
                     logger: _logger,
                     exceptionsToBeIgnored: _exceptionsToBeIgnored,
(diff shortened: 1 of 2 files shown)


iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -132,6 +133,16 @@ private async Task InitializeAndSetupClientAsync(CancellationToken cancellationT
                     cancellationToken: cancellationToken);
                 _logger.LogDebug($"The client instance has been opened.");
 
+                // The client retrieves twin values when it is initialized.
+                await RetryOperationHelper.RetryTransientExceptionsAsync(
+                    operationName: "GetTwin",
+                    asyncOperation: async () => await RetrieveTwinAsync(cancellationToken),
```
abhipsaMisra, line 139: You can update `RetryTransientExceptionsAsync` to accept `Func<Task<T>> asyncOperation` and then you can call `s_deviceClient.GetTwinAsync()` in here directly.
RetryTransientExceptionsAsync checks if the client is connected before performing the opration, so you can avoid having to add that check twice.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -132,6 +133,16 @@ private async Task InitializeAndSetupClientAsync(CancellationToken cancellationT
...
 
+                // The client retrieves twin values when it is initialized.
+                await RetryOperationHelper.RetryTransientExceptionsAsync(
+                    operationName: "GetTwin",
+                    asyncOperation: async () => await RetrieveTwinAsync(cancellationToken),
+                    shouldExecuteOperation: () => IsDeviceConnected,
+                    logger: _logger,
+                    exceptionsToBeIgnored: _exceptionsToBeIgnored,
+                    cancellationToken: cancellationToken);
+                _logger.LogDebug("The client has received twin values for the first time. ");
```
abhipsaMisra, line 144: nit: this block will be called any time the client is reinitialized and not just the first time, so let's update this log statement to something along the lines of:
_The client has retrieved the twin state after establishing connection to IoT hub_.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -74,7 +75,7 @@ public async Task RunSampleAsync(TimeSpan sampleRunningTime)
             try
             {
                 await InitializeAndSetupClientAsync(s_cancellationTokenSource.Token);
-                await Task.WhenAll(SendMessagesAsync(s_cancellationTokenSource.Token), ReceiveMessagesAsync(s_cancellationTokenSource.Token));
+                await Task.WhenAll(GetTwinUpdateAsync(s_cancellationTokenSource.Token), SendMessagesAsync(s_cancellationTokenSource.Token), ReceiveMessagesAsync(s_cancellationTokenSource.Token));
```
abhipsaMisra, line 78: We wouldn't want to call "GetTwin()" continuously on a loop, but rather only when the client is (re)connected. You can merge the logic you have in here with the "GetTwin" logic that you have in "InitializeAndSetupClientAsync".
On client (re)connect:
get twin -> compare the desired property version with last retrieved version -> if local version is less than server version - perform property update -> end

abhipsaMisra: The subsequent desired property updates would be handled by the "SetDesiredPropertyUpdateCallbackAsync" that is set in "InitializeAndSetupClientAsync".

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -219,34 +230,111 @@ private async void ConnectionStatusChangeHandler(ConnectionStatus status, Connec
...
+            { 
+                if (IsDeviceConnected)
+                {
+                    try
+                    { 
+                        twin = await s_deviceClient.GetTwinAsync();   
+                    }
+                    catch (Exception ex)
+                    { 
+                        _logger.LogError($"Unexpected error {ex}");
```
drwill-ms, line 270: Do we want to ignore these exceptions?

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -219,34 +230,111 @@ private async void ConnectionStatusChangeHandler(ConnectionStatus status, Connec
...
+                    }
+                }
+            }
+        }
+
+        // The client validates if the version of desired properties is outdated.
+        // If so, the client will "actively" get twin updates from the server.
+        private async Task GetTwinUpdateAsync (CancellationToken cancellationToken)
```
drwill-ms, line 256:
```suggestion
        private async Task GetTwinUpdateAsync(CancellationToken cancellationToken)

```

drwill-ms (changes requested):
Waiting on updates requested by Abhipsa.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
brycewang-microsoft: > You can update `RetryTransientExceptionsAsync` to accept `Func<Task<T>> asyncOperation` and then you can call `s_deviceClient.GetTwinAsync()` in here directly. RetryTransientExceptionsAsync checks if the client is connected before performing the opration, so you can avoid having to add that check twice.

modified

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
brycewang-microsoft: this method has been removed.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
brycewang-microsoft: From my tests, in the case that connection status changes from **Disconnected_Retrying** to **Connected**, twin updates on the server side during disconnection period  will not be handled by `SetDesiredPropertyUpdateCallbackAsync()` after the client gets re-connceted. That's why I wanted to call "GetTwin()" continuously on a loop. 

Meanwhile I am checking with Python and Node SDK teams as well to understand their behaviours.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -124,29 +124,55 @@ private async Task InitializeAndSetupClientAsync(CancellationToken cancellationT
...
                 // We have set the "shouldExecuteOperation" function to always try to open the connection.
                 // OpenAsync() is an idempotent call, it has the same effect if called once or multiple times on the same client.
-                await RetryOperationHelper.RetryTransientExceptionsAsync(
+                await RetryOperationHelper<Task>.RetryTransientExceptionsAsync(
                     operationName: "OpenConnection",
-                    asyncOperation: async () => await s_deviceClient.OpenAsync(cancellationToken),
+                    asyncOperation: async () => 
+                    {
+                        await s_deviceClient.OpenAsync(cancellationToken);
+                        return Task.FromResult(true);
```
drwill-ms, line 132: Because line 131 is awaited, you can just return `true`.

```suggestion
                        return true;
```

Using `Task.FromResult()` or `Task.Completed` only need to be used when a method signature indicates a `Task<>` or `Task` return type, but the method is not marked as `async` because it does not await any async methods.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -230,109 +256,39 @@ private async void ConnectionStatusChangeHandler(ConnectionStatus status, Connec
...
+                // For the purpose of this sample, we'll blindly accept all twin property write requests.
+                foreach (KeyValuePair<string, object> desiredProperty in twinUpdateRequest)
+                {
+                    _logger.LogInformation($"Setting property {desiredProperty.Key} to {desiredProperty.Value}.");
+                    reportedProperties[desiredProperty.Key] = desiredProperty.Value;
+                }
 
-            s_latestDesiredVersion = twinCollection.Version;
+                s_localDesiredPropertyVersion = twinUpdateRequest.Version;
+                _logger.LogDebug($"The desried property version on local is currently {s_localDesiredPropertyVersion}.");
```
drwill-ms, line 275:
```suggestion
                _logger.LogDebug($"The desired property version on local is currently {s_localDesiredPropertyVersion}.");

```

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -230,109 +256,39 @@ private async void ConnectionStatusChangeHandler(ConnectionStatus status, Connec
...
-                logger: _logger,
-                exceptionsToBeIgnored: _exceptionsToBeIgnored,
-                cancellationToken: cancellationToken);
+                // For the purpose of this sample, we'll blindly accept all twin property write requests.
+                await RetryOperationHelper<Task>.RetryTransientExceptionsAsync(
+                    operationName: "UpdateReportedProperties",
+                    asyncOperation: async () =>
+                    {
+                        await s_deviceClient.UpdateReportedPropertiesAsync(reportedProperties, cancellationToken);
+                        return Task.FromResult(true);
```
drwill-ms, line 283:
```suggestion
                        return true;

```

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -346,9 +302,13 @@ private async Task SendMessagesAsync(CancellationToken cancellationToken)
...
 
                     using Message message = PrepareMessage(messageCount);
-                    await RetryOperationHelper.RetryTransientExceptionsAsync(
+                    await RetryOperationHelper<Task>.RetryTransientExceptionsAsync(
                         operationName: $"SendD2CMessage_{messageCount}",
-                        asyncOperation: async () => await s_deviceClient.SendEventAsync(message),
+                        asyncOperation: async () =>
+                        {
+                            await s_deviceClient.SendEventAsync(message);
+                            return Task.FromResult(true);
```
drwill-ms, line 308:
```suggestion
                            return true;
```

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -386,9 +346,13 @@ private async Task ReceiveMessagesAsync(CancellationToken cancellationToken)
...
                     $"\nUse the IoT Hub Azure Portal or Azure IoT Explorer to send a message to this device.");
 
-                await RetryOperationHelper.RetryTransientExceptionsAsync(
+                await RetryOperationHelper<Task>.RetryTransientExceptionsAsync(
                     operationName: "ReceiveAndCompleteC2DMessage",
-                    asyncOperation: async () => await ReceiveMessageAndCompleteAsync(),
+                    asyncOperation: async () =>
+                    {
+                        await ReceiveMessageAndCompleteAsync();
+                        return Task.FromResult(true);
```
drwill-ms, line 352:
```suggestion
                        return true;
```

iot-hub/Samples/device/DeviceReconnectionSample/RetryOperationHelper.cs
```diff
@@ -27,7 +27,7 @@ internal class RetryOperationHelper
         /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
         internal static async Task RetryTransientExceptionsAsync(
             string operationName,
-            Func<Task> asyncOperation,
+            Func<Task<T>> asyncOperation,
```
drwill-ms, line 30: Why was this change to `Task<T>`? On line 39 we call it, but we don't check the result, so I wonder why `Task` alone as a return type is insufficient.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: Based on my comment in the next file, I'm thinking we shouldn't need to return true from these methods at all.

iot-hub/Samples/device/DeviceReconnectionSample/RetryOperationHelper.cs
brycewang-microsoft: I planned to call `s_deviceClient.GetTwinAsync()` as `asyncOperation` in **DeviceReconnectionSample** class, so made it with a generic return type here. But I think you are right as I don't necessarily need a value to be returned, instead I just need this value itself.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
brycewang-microsoft: corrected

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -35,7 +35,7 @@ public class DeviceReconnectionSample
         private static volatile ConnectionStatus s_connectionStatus = ConnectionStatus.Disconnected;
 
         private static CancellationTokenSource s_cancellationTokenSource;
-        private static long s_latestDesiredVersion;
+        private static long s_localDesiredPropertyVersion = 1;
```
drwill-ms, line 38:
```suggestion
        // A safe initial value for caching the twin desired properties version is 1, so the client
        // will process all previous property change requests and initialize the device application
        // after which this version will be updated to that, so we have a high water mark of which version number
        // has been processed.
        private static long s_localDesiredPropertyVersion = 1;
```

drwill-ms (approved):

PR #258, opened 2022-02-24
Move GetTwin from Client Initialization to "CONNECTED" Connection Status Change Block

brycewang-microsoft:
This can make retrieval of lost twins happen on all reconnection, as opposed to solely on (re)initialization.

### DIFF
--- a/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
+++ b/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
@@ -152,8 +152,8 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
         // It is not good practice to have async void methods, however, DeviceClient.SetConnectionStatusChangesHandler() event handler signature has a void return type.
         // As a result, any operation within this block will be executed unmonitored on another thread.
-        // To prevent multi-threaded synchronization issues, the async method InitializeClientAsync being called in here first grabs a lock
-        // before attempting to initialize or dispose the device client instance.
+        // To prevent multi-threaded synchronization issues, the async method InitializeClientAsync being called in here first grabs a lock before attempting to
+        // initialize or dispose the device client instance; the async method GetTwinAndDetectChangesAsync is implemented similarly for the same purpose.
         private async void ConnectionStatusChangeHandler(ConnectionStatus status, ConnectionStatusChangeReason reason)
         {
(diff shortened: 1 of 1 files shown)


iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -249,6 +228,34 @@ private async void ConnectionStatusChangeHandler(ConnectionStatus status, Connec
             }
         }
 
+        private async Task GetTwinOnConnectAsync(CancellationToken cancellationToken)
```
drwill-ms, line 231: I think the method name is misleading because it could be called any time. I think where you call it and a comment there is all that you need.

So, FWIW, I'd just name it `GetTwinAndDetectChanges()` or something like that, but just my opinion.

drwill-ms (approved):

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -249,6 +228,34 @@ private async void ConnectionStatusChangeHandler(ConnectionStatus status, Connec
             }
         }
 
+        private async Task GetTwinOnConnectAsync(CancellationToken cancellationToken)
+        {
```
abhipsaMisra, line 232: nit: for async methods being called from an async void callback, it is generally safer to add a lock so that we're not executing them multiple times for a single event. Eg. if multiple coe paths result in the connection status callback being triggered, we don't want each of them to be executing this block. For our specific case, ideally, the SDK shouldn't be firing multiple events for the same event change, but unless we have that explicitly tested and documented, I think it is worth handling that scenario.

iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -187,6 +163,9 @@ private async void ConnectionStatusChangeHandler(ConnectionStatus status, Connec
             {
                 case ConnectionStatus.Connected:
                     _logger.LogDebug("### The DeviceClient is CONNECTED; all operations will be carried out as normal.");
+                    // Call GetTwinOnConnectAsync() to retrieve twin values from the server once the connection status changes into Connected.
```
abhipsaMisra, line 166: nit: how about we add a new line between the log statement and the comment, for readability

abhipsaMisra (approved):

PR #259, opened 2022-02-25
Update InvokeDeviceMethod to support passing in deviceName argument

kgremban:
## Purpose

The quickstart article that references this sample (https://docs.microsoft.com/en-us/azure/iot-hub/quickstart-control-device) tells readers to name their device MyDotnetDevice or else change the hardcoded device name in the sample. This pull request updates the sample so that readers can pass in their device name as a command line argument so we don't have to enforce a specific device name. The sample remains backwards compatible, so if no deviceName argument is provided it defaults to MyDotnetDevice.

```
[ ] Yes
[ X ] No
```

## Pull Request Type

```
[ ] Bugfix
[ X ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone https://github.com/kgremban/azure-iot-samples-csharp.git
cd azure-iot-samples-csharp
git checkout feb25-qsinvokemethod
```

* Test the code
1. Use the steps in https://docs.microsoft.com/en-us/azure/iot-hub/quickstart-control-device to set up the simulated device, and then invoke its direct method.
2. Try using a device named 'MyDotnetDevice' and _not_ providing a second command line argument when invoking the direct method.
3. Try using a device with a custom name, and provide that device name as the second command line argument.

The quickstart will be updated to support custom device names after this code sample change is merged. 

## What to Check
* The InvokeDeviceMethod works whether a device name is passed to the method or not.

## Other Information
The quickstart will be updated to support custom device names after this code sample change is merged. PR for those doc changes:  https://github.com/MicrosoftDocs/azure-docs-pr/pull/189836

### DIFF
--- a/iot-hub/Quickstarts/InvokeDeviceMethod/Parameters.cs
+++ b/iot-hub/Quickstarts/InvokeDeviceMethod/Parameters.cs
@@ -9,14 +9,14 @@ namespace InvokeDeviceMethod
     internal class Parameters
     {
         [Option(
-            's',
+            'c',
             "HubConnectionString",
             HelpText = "The IoT Hub connection string. This is available under the \"Shared access policies\" in the Azure portal." +
             "\nDefaults to value of environment variable IOTHUB_CONNECTION_STRING.")]
         public string HubConnectionString { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_CONNECTION_STRING");
 
         [Option(
-            'i',
+            'd',
             "DeviceId",
             Required = false,
             HelpText = "The Id of the device to receive the direct method." +
--- a/iot-hub/Quickstarts/InvokeDeviceMethod/Parameters.cs
+++ b/iot-hub/Quickstarts/InvokeDeviceMethod/Parameters.cs
@@ -1,4 +1,5 @@
 using CommandLine;
+using System;
 
 namespace InvokeDeviceMethod
 {
@@ -9,16 +10,17 @@ namespace InvokeDeviceMethod
     {
         [Option(
             's',
-            "ServiceConnectionString",
-            Required = false,
-            HelpText = "The connection string for the IoT hub.")]
-        public string ServiceConnectionString { get; set; }
+            "HubConnectionString",
+            HelpText = "The IoT Hub connection string. This is available under the \"Shared access policies\" in the Azure portal." +
+            "\nDefaults to value of environment variable IOTHUB_CONNECTION_STRING.")]
+        public string HubConnectionString { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_CONNECTION_STRING");
 
         [Option(
-            'n',
-            "DeviceName",
+            'i',
+            "DeviceId",
             Required = false,
-            HelpText = "Name of the device to receive the direct method.")]
-        public string DeviceName { get; set; }
+            HelpText = "The Id of the device to receive the direct method." +
+            "\nDefaults to \"MyDotnetDevice\".")]
+        public string DeviceId { get; set; } = "MyDotnetDevice";
     }
 }
--- a/iot-hub/Quickstarts/InvokeDeviceMethod/Parameters.cs
+++ b/iot-hub/Quickstarts/InvokeDeviceMethod/Parameters.cs
@@ -1,0 +1,24 @@
+using CommandLine;
+
+namespace InvokeDeviceMethod
+{
+    /// <summary>
+    /// Command line parameters for the InvokeDeviceMethod sample
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            's',
+            "ServiceConnectionString",
+            Required = false,
+            HelpText = "The connection string for the IoT hub.")]
+        public string ServiceConnectionString { get; set; }
+
+        [Option(
+            'n',
+            "DeviceName",
+            Required = false,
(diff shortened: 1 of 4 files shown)


drwill-ms:
This is good stuff, Kelly. We've taken a more structured approach in our other samples and just hadn't gotten to this one yet. Would you mind following this other convention instead?

We use the [CommandLineParser](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/main/iot-hub/Samples/device/MethodSample/MethodSample.csproj#L10) package.

Initialize parameters like this: https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/main/iot-hub/Samples/device/MethodSample/Program.cs#L22-L36
Define parameters like this: https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/main/iot-hub/Samples/device/MethodSample/Parameters.cs

This would include both the service connection string and the device name.

If you wish to make environment variables an option as well, you can simply initialize the parameters.cs class properties like we do on [this sample](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/main/iot-hub/Samples/device/PnpDeviceSamples/Thermostat/Parameter.cs#L20).

kgremban:
@drwill-ms Thank you for the review and feedback. I've incorporated the new parameters structure. Since this change does break the existing quickstart, I created a new smaller PR with just this update that can quickly be published when these updates are merged: https://github.com/MicrosoftDocs/azure-docs-pr/pull/190626

iot-hub/Quickstarts/InvokeDeviceMethod/Parameters.cs
```diff
@@ -0,0 +1,24 @@
+using CommandLine;
+
+namespace InvokeDeviceMethod
+{
+    /// <summary>
+    /// Command line parameters for the InvokeDeviceMethod sample
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
```
drwill-ms, line 10: For consistency with other sample's parameters files, please match the option alias and name: https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/main/iot-hub/Samples/service/DigitalTwinClientSamples/Thermostat/Parameters.cs

iot-hub/Quickstarts/InvokeDeviceMethod/Program.cs
```diff
@@ -19,15 +20,40 @@ internal class Program
...
         
         // Connection string for your IoT Hub
-        // az iot hub show-connection-string --hub-name {your iot hub name} --policy-name service
+        // az iot hub connection-string show --policy-name service --hub-name {YourIoTHubName}
+        // Override by passing service connection string as first parameter in command line
         private static string s_connectionString = "{Your service connection string here}";
 
+        // Default device name is `MyDotnetDevice`
+        // Override by passing custom name as second parameter in command line
+        private static string s_deviceName = "MyDotnetDevice";
```
drwill-ms, line 29: Instead, you can give the parameter a default value (in the attribute, which will make that value visible when someone runs --? when running the app.

iot-hub/Quickstarts/InvokeDeviceMethod/Program.cs
```diff
@@ -19,15 +20,40 @@ internal class Program
         private static ServiceClient s_serviceClient;
         
         // Connection string for your IoT Hub
-        // az iot hub show-connection-string --hub-name {your iot hub name} --policy-name service
+        // az iot hub connection-string show --policy-name service --hub-name {YourIoTHubName}
+        // Override by passing service connection string as first parameter in command line
         private static string s_connectionString = "{Your service connection string here}";
```
drwill-ms, line 25: You can get rid of this class member too. It is not a good practice to make customers edit a code sample just to run it, so we stick with program parameters, and if desirable an environment variable option.

iot-hub/Quickstarts/InvokeDeviceMethod/Parameters.cs
kgremban: I updated the parameter names, but the shorthand flags in the other sample (-c and -d) don't work for me. I believe because they're reserved terms for dotnet already? I went with -s for string and -i for Id, but I'm happy to take suggestions for better options.

drwill-ms: You can pass [parameters to the application](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-run) and not worry about conficts with `dotnet run -- -p <cs> -d <deviceId>

kgremban: Got it. I wasn't using the `--` signifier to prefix my parameters. Now it works with -c and -d. Pushing that commit momentarily.

drwill-ms (approved):

kgremban:
@drwill-ms Thank you for your reviews and your help on this PR. What's the process for merging in this repo? Is there something I need to do to kick off the validation build that seems to be stalled?

drwill-ms:
> What's the process for merging in this repo? Is there something I need to do to kick off the validation build that seems to be stalled?

I kicked off the pipeline. I'll check it again tomorrow morning. Once it is green, if you are able to merge go ahead and do that. If not, I can merge it for you.

Good work and thanks for your patience with the changes requested!

PR #260, opened 2022-03-01
Update ConventionBasedOperations Sample

brycewang-microsoft:
## Purpose

* Get writable properties from server on CONNECT and check if local version is outdated.

PR #261, opened 2022-03-02
Update ConventionBasedOperations Sample

brycewang-microsoft:
## Purpose

* Fix merging conflicts from PR #260 
* Get writable properties from server on CONNECT and check if local version is outdated.

### DIFF
--- a/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
+++ b/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
@@ -132,7 +132,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
                             const string targetTemperatureProperty = "targetTemperature";
                             string componentName = writableProperty.Key;
 
-                            // Comparing here with Line188, the writableProperties variable is obtained by manually calling GetClientPropertiesAsync()
+                            // Comparing here with Line189, the writableProperties variable is obtained by manually calling GetClientPropertiesAsync()
                             // rather than using property update requests callback, so we are not getting a WritableClientProperty out here.
                             if (writableProperties.TryGetValue(componentName, targetTemperatureProperty, out double targetTemperatureValue))
                             { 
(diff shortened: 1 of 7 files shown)


iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -91,6 +110,70 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
+                    switch (writableProperty.Key)
+                    {
+                        case Thermostat1:
+                        case Thermostat2:
+                            const string targetTemperatureProperty = "targetTemperature";
+                            string componentName = writableProperty.Key;
+
+                            // Comparing here with Line189, the writableProperties variable is obtained by manually calling GetClientPropertiesAsync()
+                            // rather than using property update requests callback, so we are not getting a WritableClientProperty out here.
+                            if (writableProperties.TryGetValue(componentName, targetTemperatureProperty, out double targetTemperatureValue))
```
brycewang-microsoft, line 137: I don't feel confident about explanation in my comments above, so would be open to any suggestions! Same for the thermostat sample as well.

abhipsaMisra: How about:
"The `SubscribeToWritablePropertyUpdateRequestsAsync` callback makes the writable property requests available as a `WritableClientProperty`, which provides convenience methods for acknowledging the requests. Since property update requests that were potentially lost during reconnection can be retrieved only as a collection of property values, and not as a collection of `WritableClientProperty`, we will need to create the property response ack by ourselves."

@drwill-ms - thoughts?

drwill-ms (approved):

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
brycewang-microsoft: Modified.

PR #267, opened 2022-03-30
Update ConventionBasedOperations Sample

brycewang-microsoft:
Report properties with the default value and ACK when the device properties are empty.

### DIFF
--- a/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
+++ b/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
@@ -514,18 +514,18 @@ namespace Microsoft.Azure.Devices.Client.Samples
             // Check if the device properties for the current component are empty.
             if (!writableProperty.Contains(componentName) && !reportedProperty.Contains(componentName))
             {
-                await ReportDefault(componentName, TargetTemperatureProperty, cancellationToken);
+                await ReportInitialProperty(componentName, TargetTemperatureProperty, cancellationToken);
             }
         }
 
-        private async Task ReportDefault(string componentName, string propertyName, CancellationToken cancellationToken)
+        private async Task ReportInitialProperty(string componentName, string propertyName, CancellationToken cancellationToken)
         {
             var reportedProperties = new ClientPropertyCollection();
 
             // Report the default value with ACK(ac=203, av=0) as part of the PnP convention.
             // "DefaultPropertyValue" is set from the device when the desired property is not set via the hub.
             var propertyValue = _deviceClient.PayloadConvention.PayloadSerializer.CreateWritablePropertyResponse(
-                DefaultPropertyValue, ClientResponseCodes.Default, 0, DefaultACKDescription);
+                DefaultPropertyValue, ClientResponseCodes.DeviceInitialProperty, 0, DefaultACKDescription);
 
             reportedProperties.AddComponentProperty(componentName, propertyName, propertyValue);
 
--- a/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
+++ b/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
@@ -14,9 +14,11 @@ namespace Microsoft.Azure.Devices.Client.Samples
 {
     public class TemperatureControllerSample
     {
-        // The default reported "value" and "ad" for each "Thermostat" component on the client initial startup.
+        // The default reported "value", "ad" and "av" for each "Thermostat" component on the client initial startup.
+        // See https://docs.microsoft.com/en-us/azure/iot-develop/concepts-convention#writable-properties for more details in acknowledgment responses.
         private const double DefaultPropertyValue = 0d;
         private const string DefaultACKDescription = "Initialized with default value";
+        private const long DefaultACKVersion = 0L;
 
         private const string TargetTemperatureProperty = "targetTemperature";
         
@@ -525,7 +527,10 @@ namespace Microsoft.Azure.Devices.Client.Samples
             // Report the default value with ACK(ac=203, av=0) as part of the PnP convention.
             // "DefaultPropertyValue" is set from the device when the desired property is not set via the hub.
             var propertyValue = _deviceClient.PayloadConvention.PayloadSerializer.CreateWritablePropertyResponse(
-                DefaultPropertyValue, ClientResponseCodes.DeviceInitialProperty, 0, DefaultACKDescription);
+                DefaultPropertyValue, 
+                ClientResponseCodes.DeviceInitialProperty, 
+                DefaultACKVersion, 
+                DefaultACKDescription);
 
             reportedProperties.AddComponentProperty(componentName, propertyName, propertyValue);
 
--- a/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
+++ b/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
@@ -18,9 +18,8 @@ namespace Microsoft.Azure.Devices.Client.Samples
         private const string Thermostat1 = "thermostat1";
         private const string Thermostat2 = "thermostat2";
 
-        // The default reported "value" and "ad" for each "Thermostat" component.
+        // The default reported "value" for each "Thermostat" component.
         private const double defaultPropertyValue = 0d;
-        private const string defaultAckDescription = "Property set from the device without desired";
 
         private static readonly Random s_random = new Random();
         private static readonly TimeSpan s_sleepDuration = TimeSpan.FromSeconds(5);
@@ -523,14 +522,13 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
             // If the writable properties for the current component are empty, report the default value with ACK(ac=203, av=0).
             var propertyValue = _deviceClient.PayloadConvention.PayloadSerializer.CreateWritablePropertyResponse(
-                defaultPropertyValue, 203, 
-                0, defaultAckDescription);
+                defaultPropertyValue, 203, 0);
 
             reportedProperties.AddComponentProperty(componentName, propertyName, propertyValue);
 
             ClientPropertiesUpdateResponse updateResponse = await _deviceClient.UpdateClientPropertiesAsync(reportedProperties, cancellationToken);
 
-            _logger.LogDebug($"Property: Update - component=\"{componentName}\", {reportedProperties.GetSerializedString()}" +
+            _logger.LogDebug($"Report the default values for \"{componentName}\".\nProperty: Update - component=\"{componentName}\", {reportedProperties.GetSerializedString()}" +
                 $" in °C is complete with a version of {updateResponse.Version}.");
         }
     }
--- a/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
+++ b/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
@@ -121,14 +121,16 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
             if (!writableProperties.Contains(Thermostat1))
             {
-                // Report the default property "targetTemperature" with ACK for the component "thermostat1" when the writable properties are empty.
+                // Update the reported property "targetTemperature" with the default values and ACK
+                // for the component "thermostat1" when its writable properties are empty.
                 // This is a component-level property update call.
                 await UpdateReportedPropertiesWhenWritablePropertiesEmptyAsync(Thermostat1, cancellationToken);
             }
             
             if (!writableProperties.Contains(Thermostat2))
             {
-                // Report the default property "targetTemperature" with ACK for the component "thermostat2" when the writable properties are empty.
+                // Update the reported property "targetTemperature" with the default values and ACK
+                // for the component "thermostat2" when its writable properties are empty.
                 // This is a component-level property update call.
                 await UpdateReportedPropertiesWhenWritablePropertiesEmptyAsync(Thermostat2, cancellationToken);
             }
--- a/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
+++ b/iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
@@ -15,11 +15,24 @@ namespace Microsoft.Azure.Devices.Client.Samples
 {
     public class TemperatureControllerSample
     {
+        // The specific response codes in this sample.
+        public class ClientResponseCodes : CommonClientResponseCodes
+        {
+            /// <summary>
+            /// This code does not comply with HTTP semantics. Instead, it indicates that the device client
+            /// reports the default value with ACK when its properties are empty.
+            /// </summary>
+            public const int Default = 203;
+        }
+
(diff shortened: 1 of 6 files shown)


iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -496,5 +515,23 @@ private static double GenerateTemperatureWithinRange(int max = 50, int min = 0)
...
+
+        private async Task UpdateReportedPropertiesWhenWritablePropertiesEmptyAsync(string componentName, CancellationToken cancellationToken)
+        {
+            const string propertyName = "targetTemperature";
+            
+            var reportedProperties = new ClientPropertyCollection();
+
+            // If the writable properties for the current component are empty, report the default value with ACK(ac=203, av=0).
+            var propertyValue = _deviceClient.PayloadConvention.PayloadSerializer.CreateWritablePropertyResponse(
+                defaultPropertyValue, 203, 0);
```
brycewang-microsoft, line 527: `av` is 0, representing that the acknowledgment version is not referring to any desired properties as they are empty now.

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -18,6 +18,9 @@ public class TemperatureControllerSample
         private const string Thermostat1 = "thermostat1";
         private const string Thermostat2 = "thermostat2";
 
+        // The default reported "value" for each "Thermostat" component.
+        private const double defaultPropertyValue = 0d;
```
azabbasi, line 22: constants can be named with sentence casing
`DefaultPropertyValue`

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -496,5 +515,23 @@ private static double GenerateTemperatureWithinRange(int max = 50, int min = 0)
         {
             return Math.Round(s_random.NextDouble() * (max - min) + min, 1);
         }
+
+        private async Task UpdateReportedPropertiesWhenWritablePropertiesEmptyAsync(string componentName, CancellationToken cancellationToken)
```
azabbasi, line 519: UpdateEmptyReportedProperties ? 
This method name is a bit too long.

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
azabbasi: Can we describe what 203 and 0 represent and the reasoning behind the ACK value we have chosen?

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
brycewang-microsoft: > This method name is a bit too long. `RespondToEmptyWritableProperty` maybe? thoughts?

Sure, I can shorten this and other similar names. What about `UpdateForEmptyWritableProperty` as we don't actually respond to anything here?

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
brycewang-microsoft: > Can we describe what 203 and 0 represent and the reasoning behind the ACK value we have chosen?

0 represents that `av` is not referring to any desired properties set via the hub. When a desired property is set up, its version number starts from 1 by default. 
For 203, I can check with @rido-min for the reasoning though, and add comments into our samples further so it can be clearer to our users.

rido-min: we are describing these cases in the PnP convention

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
azabbasi: well, technically you are responding by updating your reported properties. It's considered a response to a writable property request.

azabbasi (approved):

iot-hub/Samples/device/ConventionBasedOperations/Thermostat/ThermostatSample.cs
```diff
@@ -83,12 +86,18 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
             }
         }
 
-        private async Task GetWritablePropertiesAndHandleChangesAsync()
+        private async Task GetWritablePropertiesAndHandleChangesAsync(CancellationToken cancellationToken)
```
rido-min, line 89: have you considered adding the property name as a parameter? it will be easier to reuse this method if there is more than one property to initialize

iot-hub/Samples/device/ConventionBasedOperations/Thermostat/ThermostatSample.cs
```diff
@@ -277,5 +286,24 @@ private static double GenerateTemperatureWithinRange(int max = 50, int min = 0)
...
+        private async Task UpdateForEmptyWritableProperty(CancellationToken cancellationToken)
+        {
+            const string propertyName = "targetTemperature";
+
+            var reportedProperties = new ClientPropertyCollection();
+
+            // If the writable properties are empty, report the default value with ACK(ac=203, av=0) as part of the PnP convention.
+            // "DefaultPropertyValue" is set from the device when the desired property is not set via the hub.
+            var propertyValue = _deviceClient.PayloadConvention.PayloadSerializer.CreateWritablePropertyResponse(
+                DefaultPropertyValue, 203, 0);
```
rido-min, line 299: can we add the 203 to `CommonClientResponseCodes` ?

iot-hub/Samples/device/ConventionBasedOperations/Thermostat/ThermostatSample.cs
```diff
@@ -277,5 +286,24 @@ private static double GenerateTemperatureWithinRange(int max = 50, int min = 0)
         {
             return Math.Round(s_random.NextDouble() * (max - min) + min, 1);
         }
+
+        private async Task UpdateForEmptyWritableProperty(CancellationToken cancellationToken)
```
rido-min, line 290: consider passing a param with the value to initialize

rido-min (commented):
added some comments to Thermostat, should be applicable to TemperatureController

iot-hub/Samples/device/ConventionBasedOperations/Thermostat/ThermostatSample.cs
brycewang-microsoft: `GetWritablePropertiesAndHandleChangesAsync` is called only on `ConnectionStatus.Connected`, so it is more reasonable to set "propertyName"/ "componentName" as parameter for the method `UpdateForEmptyWritableProperty` which is implemented in TemperatureController sample. I can do the same for Thermostat though.

iot-hub/Samples/device/ConventionBasedOperations/Thermostat/ThermostatSample.cs
brycewang-microsoft: added

iot-hub/Samples/device/ConventionBasedOperations/Thermostat/ThermostatSample.cs
brycewang-microsoft: We may leave "203" as in the samples for now, and change it to something like `CommonClientResponseCodes.Default` after we update status codes in the class `CommonClientResponseCodes`. Are we considering to add other codes [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/preview/shared/src/CommonClientResponseCodes.cs)?

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -110,12 +113,28 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
             ClientPropertyCollection writableProperties = properties.WritablePropertyRequests;
             long serverWritablePropertiesVersion = writableProperties.Version;
 
+            if (!writableProperties.Contains(Thermostat1))
+            {
+                // Update the reported property "targetTemperature" with the default values and ACK
+                // for the component "thermostat1" when its writable properties are empty.
+                // This is a component-level property update call.
+                await UpdateForEmptyWritableProperty(Thermostat1, cancellationToken);
+            }
```
abhipsaMisra, line 128: If I'm understanding this correctly, the intention here is to update the "reported" property value to some default value on initial start-up. If that is correct, then you don't need to call this from within the status change callback but rather only on the first initial client set up. If you call this from within the callback, then anytime the client reconnects the reported property value will be reset to the previously defined default value. Is that the intention here?

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -496,5 +515,24 @@ private static double GenerateTemperatureWithinRange(int max = 50, int min = 0)
...
+        private async Task UpdateForEmptyWritableProperty(string componentName, CancellationToken cancellationToken)
+        {
+            const string propertyName = "targetTemperature";
+            
+            var reportedProperties = new ClientPropertyCollection();
+
+            // If the writable properties are empty, report the default value with ACK(ac=203, av=0) as part of the PnP convention.
+            // "DefaultPropertyValue" is set from the device when the desired property is not set via the hub.
+            var propertyValue = _deviceClient.PayloadConvention.PayloadSerializer.CreateWritablePropertyResponse(
+                DefaultPropertyValue, 203, 0);
```
abhipsaMisra, line 528: Instead of passing in the status code directly, I'd recommend extending `CommonClientResponseCodes` and adding in the application specific codes with the doc comments. This will guide users to adopt the recommended practices.

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
brycewang-microsoft: @rido-min Do we want to report the default values for empty writable properties on **initial start-up only** or **each connection/reconnection** ?

rido-min: I'd say that defaultValues should be set only on *initial startup*, the problem is that during reconnection we don't know if the defaultValue has been successfully accepted. 

Ideally, we should get the twin on each connection/reconnection, compare the reported properties, and if they match the current values we should not write those again. 

To answer @abhipsaMisra question. The default values only make sense if the twin was empty (no previous reported/desired values found)

rido-min:
Can you add a sample description `ad` when updating the default value?

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
brycewang-microsoft: In such a case I will move the report-default-values call to the client initialization instead.

brycewang-microsoft:
> Can you add a sample description `ad` when updating the default value?

Sure. As the description is not on the MS docs yet, can you share the `ad` contents here?

rido-min:
'ad' = "targetTemperature initialized with default value"

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -496,5 +515,24 @@ private static double GenerateTemperatureWithinRange(int max = 50, int min = 0)
         {
             return Math.Round(s_random.NextDouble() * (max - min) + min, 1);
         }
+
+        private async Task UpdateForEmptyWritableProperty(string componentName, CancellationToken cancellationToken)
```
azabbasi, line 519: Maybe we should make this method more generic and not hard code the property name in the method, like passing the property name as a parameter?

brycewang-microsoft: sure! Actually I am re-constructing this method, and will bring this to the new one.

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -72,6 +88,11 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
             _logger.LogDebug($"Subscribe to commands.");
             await _deviceClient.SubscribeToCommandsAsync(HandleCommandsAsync, cancellationToken);
 
+            // For each component, check if the corresponding properties (both writable and reported) are empty on the initial startup.
+            // If so, report the default values with ACK to the hub.
+            await CheckEmptyProperties(Thermostat1, cancellationToken);
```
brycewang-microsoft, line 93: Call `CheckEmptyProperties()` after setting all handlers; otherwise (call this method before `SetConnectionStatusChangesHandler`), reporting default values by calling `CheckEmptyProperties()` will change the connection status into "Connected", so all existing writable properties will not be obtained by `GetWritablePropertiesAndHandleChangesAsync()` and reported to the hub further.

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -15,6 +15,22 @@ namespace Microsoft.Azure.Devices.Client.Samples
 {
     public class TemperatureControllerSample
     {
+        // The specific response codes in this sample.
+        public class ClientResponseCodes : CommonClientResponseCodes
```
abhipsaMisra, line 19: nit: I'd separate this out into its own class file so that it is easily discoverable.

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -15,6 +15,22 @@ namespace Microsoft.Azure.Devices.Client.Samples
...
     public class TemperatureControllerSample
     {
+        // The specific response codes in this sample.
+        public class ClientResponseCodes : CommonClientResponseCodes
+        {
+            /// <summary>
+            /// This code does not comply with HTTP semantics. Instead, it indicates that the device client
+            /// reports the default value with ACK when its properties are empty.
+            /// </summary>
+            public const int Default = 203;
```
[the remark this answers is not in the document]
abhipsaMisra: Since this status code is specific only to twin initial setup, I'd consider renaming it from "default" to "TwinInitialValue". Thoughts?

abhipsaMisra (approved):

iot-hub/Samples/device/ConventionBasedOperations/TemperatureController/TemperatureControllerSample.cs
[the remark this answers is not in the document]
brycewang-microsoft: @rido-min do we have a preference for the name of code `203` here? something like "TwinInitialValue"?

brycewang-microsoft:
Leave this as a draft while waiting for updates on Azure docs.

PR #286, opened 2022-06-23
Sample for DPS trust bundle

abhipsaMisra:
Readme: https://github.com/Azure-Samples/azure-iot-samples-csharp/tree/abmisr/trustBundle/provisioning/Samples/device/SymmetricKeyWithTrustBundleSample


provisioning/Samples/device/SymmetricKeyWithTrustBundleSample/TrustBundleSample.cs
```diff
@@ -0,0 +1,113 @@
...
+using System;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Samples
+{
+    /// <summary>
+    ///
```
timtay-microsoft, line 17: Intentionally empty?

provisioning/Samples/device/SymmetricKeyWithTrustBundleSample/readme.md
```diff
@@ -0,0 +1,132 @@
...
+}
+```
+
+## Sample
+
+The `SymmetricKeyWithTrustBundleSample` demonstrates how to use the public API to make a registration request to DPS and receive the trust bundle certificates in return.
+
+1. Prerequisites
+
+    1. Using curl, submit the following command to the DPS Service API to create the Trust Bundle in DPS. This command uploads a test private root certificate to DPS as a trust bundle.
```
timtay-microsoft, line 59: Is this step doable using the provisioning service client? Asking users to use curl to create a trust bundle feels a bit odd

timtay-microsoft (approved):

provisioning/Samples/device/SymmetricKeyWithTrustBundleSample/TrustBundleSample.cs
abhipsaMisra: Good catch, I'll add a comment in here

provisioning/Samples/device/SymmetricKeyWithTrustBundleSample/readme.md
abhipsaMisra: This isn't available on the provisioning client. From the look of it, I think this might be a control plane operation since it has to do with updating the provisioning service instance itself (similar to how we upload and verify the root certificates for group enrollments on the provisioning service).
Portal and CLI support are expected to be added later on. For now, these curl commands are all that are available.

PR #296, opened 2022-08-02
Update import/export sample

drwill-ms:
Update to latest SDK version.
Add parameters file and command-line parsing.
Fix build warnings.
Update to newer C# lang.

### DIFF
--- a/iot-hub/Samples/service/FileUploadNotificationReceiverSample/FileUploadNotificationReceiverSample.csproj
+++ b/iot-hub/Samples/service/FileUploadNotificationReceiverSample/FileUploadNotificationReceiverSample.csproj
@@ -3,12 +3,12 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
-    <LangVersion>8</LangVersion>
+    <LangVersion>9.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Devices" Version="1.36.0" />
-    <PackageReference Include="CommandLineParser" Version="2.8.0" />
+    <PackageReference Include="CommandLineParser" Version="2.9.1" />
     <PackageReference Include="System.Text.Json" Version="6.0.1" />
   </ItemGroup>
   <ItemGroup>
--- a/iot-hub/Samples/service/ImportExportDevicesSample/ImportExportDevicesSample.cs
+++ b/iot-hub/Samples/service/ImportExportDevicesSample/ImportExportDevicesSample.cs
@@ -1,16 +1,18 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-using Microsoft.Azure.Devices.Common;
-using Microsoft.Azure.Storage;
-using Microsoft.Azure.Storage.Blob;
-using Microsoft.Azure.Storage.RetryPolicies;
-using Newtonsoft.Json;
+
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.Azure.Storage;
+using Microsoft.Azure.Storage.Blob;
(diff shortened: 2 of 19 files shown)


iot-hub/Samples/service/ImportExportDevicesSample/ImportExportDevicesSample.cs
```diff
@@ -44,37 +46,36 @@ public class ImportExportDevicesSample
...
+            int devicesToAdd,
+            bool shouldCopyDevices,
+            bool shouldDeleteSourceDevices,
+            bool shouldDeleteDestDevices)
         {
             // This sets cloud blob container and returns container uri (w/shared access token).
             _containerUriWithSas = await PrepareStorageForImportExportAsync(_storageAccountConnectionString).ConfigureAwait(false);
 
-            if (addDevices)
+            if (devicesToAdd > 0)
```
drwill-ms, line 57: We don't need the `addDevices` bool parameter if we can simply use the number of devices to add being greater than 0 to decide whether or not to do it.

iot-hub/Samples/service/ImportExportDevicesSample/ImportExportDevicesSample.cs
```diff
@@ -273,8 +273,8 @@ private async Task GenerateAndAddDevices(string hubConnectionString, string cont
                     {
                         SymmetricKey = new SymmetricKey
                         {
-                            PrimaryKey = CryptoKeyGenerator.GenerateKey(32),
-                            SecondaryKey = CryptoKeyGenerator.GenerateKey(32),
+                            PrimaryKey = GenerateKey(32),
```
drwill-ms, line 276: We've deprecateds the method previous used in the SDK, so I added a helper method below to do this.

iot-hub/Samples/service/ImportExportDevicesSample/Program.cs
```diff
@@ -4,192 +4,53 @@
...
-            Console.WriteLine($"  num to add = {_numToAdd}.");
-            Console.WriteLine($"  copy devices = {_copyDevices}.");
-            Console.WriteLine($"  delete source devices = {_deleteSourceDevices}.");
-            Console.WriteLine($"  delete dest devices = {_deleteDestDevices}.");
-            Console.WriteLine($"  IoTHubConnString = '{_envIotHubConnectionString}'.");
-            Console.WriteLine($"  IoTHubDestString = '{_envDestIotHubConnectionString}'.");
-            Console.WriteLine($"  storage connection string  = '{_envStorageAccountConnectionString}'.");
+            // Parse application parameters
+            Parameters parameters = null;
+            ParserResult<Parameters> result = Parser.Default.ParseArguments<Parameters>(args)
```
drwill-ms, line 28: Now with structured parameter parsing!

brycewang-microsoft (approved):

iot-hub/Samples/service/FileUploadNotificationReceiverSample/FileUploadNotificationReceiverSample.csproj
```diff
@@ -3,12 +3,12 @@
...
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
-    <LangVersion>8</LangVersion>
+    <LangVersion>9.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Devices" Version="1.36.0" />
-    <PackageReference Include="CommandLineParser" Version="2.8.0" />
+    <PackageReference Include="CommandLineParser" Version="2.9.1" />
```
timtay-microsoft, line 11: Is there a good way to factor this out into a variable in .NET? In Java, we have all dependency versions like these centralized and it has made dependency management much easier

timtay-microsoft: https://github.com/Azure/azure-iot-sdk-java/blob/edb759adc76972b407cd64ab6aae0549895af094/pom.xml#L22

drwill-ms: It is possible, but makes the samples less readable and then it doesn't work with the NuGet package manager anymore. Not worth it, IMHO.

PR #298, opened 2022-08-03
Update to latest storage SDK for import/export sample

drwill-ms:


iot-hub/Samples/service/ImportExportDevicesSample/ImportExportDevicesSample.cs
```diff
@@ -50,31 +51,31 @@ public class ImportExportDevicesSample
             bool shouldDeleteSourceDevices,
             bool shouldDeleteDestDevices)
         {
-            // This sets cloud blob container and returns container uri (w/shared access token).
-            _containerUriWithSas = await PrepareStorageForImportExportAsync(_storageAccountConnectionString).ConfigureAwait(false);
```
drwill-ms, line 54: Samples shouldn't use `ConfigureAwait(false)` given we don't know the context customers may copy/paste this code into and what problems that might cause. This is Track2 guidance and we've brought it to our Track1 samples, when we notice it.

timtay-microsoft (approved):

brycewang-microsoft (approved):

PR #301, opened 2022-08-12
Improved nightly clean up task with clean only build option

andyk-ms:
Add clear switch to build script for clean up only run option
Run 3 clean up processes, skip all samples runs
Shorten execution time from 40+ minutes to about 5 minutes

### DIFF
--- a/vsts/vsts.yaml
+++ b/vsts/vsts.yaml
@@ -37,7 +37,7 @@ jobs:
           version: 2.1.x
           performMultiLevelLookup: true
           installationPath: $(Agent.ToolsDirectory)/dotnet
-      - powershell: ./build.ps1 -clean -configuration Release -build -run
+      - powershell: ./build.ps1 -clean -configuration Release -build -clear
         displayName: build and run
         env:
           # Environment variables for IoT hub device samples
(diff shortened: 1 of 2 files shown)


vsts/vsts.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
           version: 2.1.x
           performMultiLevelLookup: true
           installationPath: $(Agent.ToolsDirectory)/dotnet
-      - powershell: ./build.ps1 -clean -configuration Release -build -run
+      - powershell: ./build.ps1 -clean -configuration Release -build -clear
```
abhipsaMisra, line 40: We'll still want to have a pipeline that runs all our samples. I think we should split out just the sample cleanup part to a different yaml and have that run independent of the entire sample run.

Once we move our samples out of this repo and into our sourcecode repo, we can merge the sample nightly run with our E2E nightly run.

abhipsaMisra (approved):

vsts/vsts.yaml
andyk-ms: Agree. Eventually, I think we don't have this cleanup pipeline, test will clean up for its own test data and config.

PR #314, opened 2022-08-25
parameterized JobsSample

patilsnr:
## Purpose

* Parameterizes sample with command line arguements/parser
* Implements suggested changes by IDE

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Code style update (formatting, local variables)
[ ] Other... Please describe:
```

### DIFF
--- a/iot-hub/Samples/service/JobsSample/JobsSample.cs
+++ b/iot-hub/Samples/service/JobsSample/JobsSample.cs
@@ -57,7 +57,7 @@ namespace Microsoft.Azure.Devices.Samples
             // *************************************** Get all Jobs ***************************************
             IEnumerable<JobResponse> queryResults = await _jobClient.CreateQuery().GetNextAsJobResponseAsync().ConfigureAwait(false);
 
-            List<JobResponse> getJobs = queryResults.ToList();
+            var getJobs = queryResults.ToList();
             Console.WriteLine($"getJobs return {getJobs.Count} result(s)");
 
             foreach (JobResponse job in getJobs)
--- a/iot-hub/Samples/service/JobsSample/JobsSample.cs
+++ b/iot-hub/Samples/service/JobsSample/JobsSample.cs
@@ -8,7 +8,7 @@ using System.Text.Json;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Shared;
 
-namespace Microsoft.Azure.Devices.Samples
+namespace Microsoft.Azure.Devices.Samples.JobsSample
 {
     public class JobsSample
     {
@@ -57,7 +57,7 @@ namespace Microsoft.Azure.Devices.Samples
             // *************************************** Get all Jobs ***************************************
             IEnumerable<JobResponse> queryResults = await _jobClient.CreateQuery().GetNextAsJobResponseAsync().ConfigureAwait(false);
 
-            var getJobs = queryResults.ToList();
+            List<JobResponse> getJobs = queryResults.ToList();
             Console.WriteLine($"getJobs return {getJobs.Count} result(s)");
 
             foreach (JobResponse job in getJobs)
--- a/iot-hub/Samples/service/JobsSample/Parameters.cs
+++ b/iot-hub/Samples/service/JobsSample/Parameters.cs
@@ -6,7 +6,7 @@ namespace Microsoft.Azure.Devices.Samples.JobsSample
     /// <summary>
     /// Parameters for the application.
     /// </summary>
-    ///     /// <remarks>
+    /// <remarks>
     /// To get these connection strings, log into https://portal.azure.com, go to Resources, open the IoT hub, open Shared Access Policies, open iothubowner, and copy a connection string.
     /// </remarks>
     internal class Parameters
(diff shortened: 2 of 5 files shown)


iot-hub/Samples/service/JobsSample/JobsSample.cs
```diff
@@ -57,7 +57,7 @@ public async Task RunSampleAsync()
             // *************************************** Get all Jobs ***************************************
             IEnumerable<JobResponse> queryResults = await _jobClient.CreateQuery().GetNextAsJobResponseAsync().ConfigureAwait(false);
 
-            List<JobResponse> getJobs = queryResults.ToList();
+            var getJobs = queryResults.ToList();
```
drwill-ms, line 60: Type not evident from a call to `ToList()`. What type of elements are in the list? Keep the explicit type.

iot-hub/Samples/service/JobsSample/Parameters.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace JobsSample
+{
+    /// <summary>
+    /// Parameters for the application.
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            'c',
+            "HubConnectionString",
```
drwill-ms, line 13: Did we decide to go with `c` everywhere for connection string? If not, then we stick with `p`.

iot-hub/Samples/service/JobsSample/Parameters.cs
```diff
@@ -0,0 +1,18 @@
...
+            Required = true,
+            HelpText = "The connection string of the IoT Hub instance to connect to.")]
```
drwill-ms, line 15:
```suggestion
            HelpText = "The connection string of the IoT hub instance to connect to.")]
```

drwill-ms, line 15: Is this a service connection string? If so, which one is required? There are several different built-in types with various permissions.

iot-hub/Samples/service/JobsSample/Program.cs
```diff
@@ -3,27 +3,28 @@
 
 using System;
 using System.Threading.Tasks;
+using CommandLine;
+using JobsSample;
 
 namespace Microsoft.Azure.Devices.Samples
```
drwill-ms, line 9: Change this namespace to `JobSample` and you won't need the `using` statement on 7.

iot-hub/Samples/service/JobsSample/Parameters.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Parameters for the application.
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            'c',
+            "HubConnectionString",
+            Required = true,
```
drwill-ms, line 14: Can't be required if we also take it via environment variable.

drwill-ms (changes requested):
Misc comments inline.

iot-hub/Samples/service/JobsSample/Parameters.cs
patilsnr: Do we want `c` to be specific for a Hub Connection string and `p` to be specific for a device connection string? That would make sense to me but everything could also just be `p` as well

iot-hub/Samples/service/JobsSample/Parameters.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    /// Parameters for the application.
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            'p',
+            "HubConnectionString",
+            Required = false,
+            HelpText = "The connection string of the IoT hub instance to connect to. This can be located under \"Shared Access Policies\" in the Iot hub.")]
```
drwill-ms, line 15: You can move the "This can be located under..." up to class remarks, per my PR.

https://github.com/Azure-Samples/azure-iot-samples-csharp/pull/316

drwill-ms: Also, add a note that it defaults to the env var, by name. Also demonstrated in that PR.

iot-hub/Samples/service/JobsSample/Parameters.cs
```diff
@@ -0,0 +1,18 @@
...
+        public string HubConnectionString { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_CONNECTION_STRING");
+    }
```
drwill-ms, line 17: Needs a validate method, and that called from Program.cs. See PR.

iot-hub/Samples/service/JobsSample/Parameters.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using CommandLine;
+
+namespace Microsoft.Azure.Devices.Samples.JobsSample
+{
+    /// <summary>
+    /// Parameters for the application.
+    /// </summary>
+    ///     /// <remarks>
```
drwill-ms, line 9: fix

drwill-ms (approved):

PR #323, opened 2022-08-31
Gracefully handle throttling in sample pipelines

patilsnr:


iot-hub/Samples/service/JobsSample/JobsSample.cs
```diff
@@ -40,50 +41,58 @@ public async Task RunSampleAsync()
...
-            Console.WriteLine($"Schedule twin job {jobId} for {DeviceId}...");
-            JobResponse createJobResponse = await _jobClient
-                .ScheduleTwinUpdateAsync(
-                    jobId,
-                    query,
-                    twin,
-                    DateTime.UtcNow,
-                    (long)TimeSpan.FromMinutes(2).TotalSeconds);
-
-            Console.WriteLine("Schedule response");
```
timtay-microsoft, line 52: There is no need for this line and below to be in the try/catch block since they won't throw a ThrottlingException.

Try something like

```csharp

JobResponse createJobResponse;

try
{
            createJobResponse = await _jobClient
                .ScheduleTwinUpdateAsync(
                    jobId,
                    query,
                    twin,
                    DateTime.UtcNow,
                    (long)TimeSpan.FromMinutes(2).TotalSeconds);
}
catch (ThrottlingException)
{
    Console.WriteLine("...");
    return;
}

Console.WriteLine("Schedule response");
```

patilsnr: awesome, I'll get on changing this and moving it to other places where it's needed. Thanks Tim!

iot-hub/Samples/service/AutomaticDeviceManagementSample/AutomaticDeviceManagementSample.cs
```diff
@@ -90,24 +99,40 @@ private static void CreateMetricsAndTargetCondition(Configuration configuration)
 
         private async Task DeleteConfigurationAsync(string configurationId)
         {
-            await _registryManager.RemoveConfigurationAsync(configurationId);
+            try
+            {
+                await _registryManager.RemoveConfigurationAsync(configurationId);
```
timtay-microsoft, line 104: This request won't hit job throttling limits since it is not a job request, so you can skip the try/catch here

iot-hub/Samples/service/AutomaticDeviceManagementSample/AutomaticDeviceManagementSample.cs
```diff
@@ -90,24 +99,40 @@ private static void CreateMetricsAndTargetCondition(Configuration configuration)
...
+                    string configurationString = JsonConvert.SerializeObject(configuration, Formatting.Indented);
+                    Console.WriteLine(configurationString);
+                    Thread.Sleep(1000);
+                }
 
-            // Check configuration's metrics for expected conditions
-            foreach (Configuration configuration in configurations)
+                Console.WriteLine("Configurations received");
+            }
+            catch (ThrottlingException)
```
timtay-microsoft, line 131: There doesn't appear to be a service request in the try block, so no ThrottlingException will ever be thrown. Because of that, you can remove the try/catch here

iot-hub/Samples/service/JobsSample/JobsSample.cs
```diff
@@ -40,18 +41,26 @@ public async Task RunSampleAsync()
...
+                        query,
+                        twin,
+                        DateTime.UtcNow,
+                        (long)TimeSpan.FromMinutes(2).TotalSeconds);
+
+                Console.WriteLine("Schedule response");
+                Console.WriteLine(JsonSerializer.Serialize(createJobResponse, new JsonSerializerOptions { WriteIndented = true }));
+                Console.WriteLine();
+            }
+            catch (ThrottlingException)
```
brycewang-microsoft, line 59: How about adding some doc reference in the catch block in case that the user would like to learn more about the quota?
https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-quotas-throttling#other-limits

brycewang-microsoft (approved):

iot-hub/Samples/service/JobsSample/JobsSample.cs
timtay-microsoft: +1 to this. Whatever it took to help you understand this exception should be written here so that other users can learn about it as well

timtay-microsoft (approved):

PR #326, opened 2022-09-02
Updated with parameters for config

andyk-ms:
Eliminated pipeline variable
Updated with parameters on run config, allow manual run with clearing service resources only (no test)
Default with full run from scheduled trigger

### DIFF
--- a/vsts/vsts.yaml
+++ b/vsts/vsts.yaml
@@ -24,7 +24,7 @@ resources:
 jobs:
   ### Linux build ###
   - job: LINUX
-    displayName: Linux (runs ~${{ parameters.jobTimeoutInMinutes }}mins)
+    displayName: Linux (timeout=${{ parameters.jobTimeoutInMinutes }}mins)
     condition: succeeded()
     pool:
       vmImage: 'ubuntu-latest'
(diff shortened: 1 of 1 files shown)


vsts/vsts.yaml
```diff
@@ -1,17 +1,34 @@
...
+
 resources:
   - repo: self
     clean: true
 
 jobs:
   ### Linux build ###
   - job: LINUX
-    displayName: Linux
+    displayName: Linux (timeout=${{ parameters.jobTimeoutInMinutes }}mins)
```
drwill-ms, line 27: Seems odd to make that one property part of the title.

drwill-ms: If we need this printed somewhere, I'd just document it in a step.

drwill-ms (approved):

vsts/vsts.yaml
andyk-ms: I thought it allows us to see if the parameter is set, seeing the timeout knowing if custom parameter is taken on the job.
Alternative is to print this out on the log. We can change it if this info is not helpful.

andyk-ms:
Updated the pipeline yaml with parameters.

